PR #40, opened 2014-12-14
Fixed a potential namespace conflicting issue.

ronxin:
On my Ubuntu system, there is namespace conflicting issue caused by uintmax_t declared in both boost and std namespaces (possibly due to my outdated boost library include files). Avoiding using "using namespace **" fixed the problem.


a1k0n:
this is also just good style, thank you

Issue #3, opened 2014-01-03
annoy compilation fail


piskvorky:
Installing annoy from repo on EC2 m3.2xlarge instance (Ubuntu 13.10):

```
$ uname -a
Linux m3-2xlarge 3.11.0-12-generic #19-Ubuntu SMP Wed Oct 9 16:20:46 UTC 2013 x86_64 x86_64 x86_64 GNU/Linux

$ gcc --version
gcc (Ubuntu/Linaro 4.8.1-10ubuntu9) 4.8.1
```

fails with

```
$ python setup.py build
running build
running build_py
creating build
creating build/lib.linux-x86_64-2.7
creating build/lib.linux-x86_64-2.7/annoy
copying annoy/__init__.py -> build/lib.linux-x86_64-2.7/annoy
running build_ext
building 'annoy.annoylib' extension
creating build/temp.linux-x86_64-2.7
creating build/temp.linux-x86_64-2.7/src
x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I/usr/include/python2.7 -c src/annoylib.cc -o build/temp.linux-x86_64-2.7/src/annoylib.o
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++ [enabled by default]
src/annoylib.cc: In instantiation of ‘void AnnoyIndexPython<T, Distance>::add_item_py(int, const boost::python::list&) [with T = float; Distance = Angular<float>]’:
src/annoylib.cc:542:30:   required from ‘void expose_methods(boost::python::class_<C>) [with C = AnnoyIndexPython<float, Angular<float> >]’
src/annoylib.cc:555:117:   required from here
src/annoylib.cc:508:25: error: ‘add_item’ was not declared in this scope, and no declarations were found by argument-dependent lookup at the point of instantiation [-fpermissive]
     add_item(item, &w[0]);
                         ^
src/annoylib.cc:508:25: note: declarations in dependent base ‘AnnoyIndex<float, Angular<float> >’ are not found by unqualified lookup
src/annoylib.cc:508:25: note: use ‘this->add_item’ instead
src/annoylib.cc: In instantiation of ‘boost::python::list AnnoyIndexPython<T, Distance>::get_nns_by_vector_py(boost::python::list, size_t) [with T = float; Distance = Angular<float>; size_t = long unsigned int]’:
[...]

Issue #1, opened 2013-10-18
return value of fopen() in save() is not checked


a1k0n:
segfaults if you try to write somewhere you can't.

PR #42, opened 2015-02-26
inremental add_item_to_index and load_memory

LongbinChen:
the main changes:

1. to avoid all the mass around mmap, i added a function, load_memory(), to load everything into the memory. The following operations would work on memory to avoid complications on mmap. Believe me I had tried to use stuffs like mremap and struggled a bit and feel we better off to load it memory and save to a different filename. 

2. the function add_item_to_index(T* w) would add a new item to existing index.  The newly added item will be saved at the space right after the last existing item. Before doing so, of course, I have to move the nodes at the position to the end of the array -- again, i have to remove the roots at the end of array when I load the file. 

3. also, in order to be able to move nodes around, i have to add an additional field "parent" to save a node's parents. 

4. i will wrap a python service ( probably based on tornado) to this index. so when there is a request to add an item, the service will load the index in memory, add the item, and save the updated index to a different file. the service keep a point to "current" index and "working" index. when serving traffic, the request still read from mmap.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,11 +31,13 @@ Summary of features
 -------------------
 
 * Euclidean distance (squared) or cosine similarity (using the squared distance of the normalized vectors)
-* Works better if you don't have too many dimensions (like <100) but seems to perform surprisingly well even up to 1,000 dimensions
-* Small memory usage
+* Works better if you don't have too many dimensions (like <100) but seems to perform surprisingly well even up to 1,000 dimensions; works pretty fine with dimension up to 4000. 
+* Small memory usage in searching because memory map is used
 * Lets you share memory between multiple processes
-* Index creation is separate from lookup (in particular you can not add more items once the tree has been created)
+* Index creation is separate from lookup
 * Native Python support
+* Support dynamic index add. ( will support delete and update in the future)
+* Each sample contains a field label for filtering results -- i.e. search query can contain a list of target labels to search for.
 
 Code example
 ____________
@@ -48,7 +50,7 @@ ____________
       v = []
       for z in xrange(f):
           v.append(random.gauss(0, 1))
-      t.add_item(i, v, [1])
+      t.add_item(i, v, 1)  # add items with label 1
 
   t.build(10) # 10 trees
   t.save('test.tree')
@@ -57,7 +59,7 @@ ____________
 
   u = AnnoyIndex(f)
   u.load('test.tree') # super fast, will just mmap the file
-  print(u.get_nns_by_item(0, 1000, [1], 10000)) # will find the 1000 nearest neighbors
+  print(u.get_nns_by_item(0, 1000, [1,2], 10000)) # will find the 1000 nearest neighbors, accurate search from top 10000 candidates with label 1 or 2
 
 
 Right now it only accepts integers as identifiers for items. Note that it will allocate memory for max(id)+1 items because it assumes your items are numbered 0 … n-1. If you need other id's, you will have to keep track of a map yourself.
--- a/README.rst
+++ b/README.rst
@@ -8,6 +8,15 @@ Annoy
 What is this?
 -------------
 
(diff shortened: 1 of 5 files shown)


erikbern:
Let me take a look at this in the next few days

erikbern:
On #1: I think *the* key feature of Annoy is that it uses mmap rather than allocates memory from the heap. This will not change. So please if you can figure out another way around it. One way is to keep the original chunk in mmap and use heap memory for any new nodes.

erikbern:
On #2: You don't *have* to move the root nodes at the end of the index. It's nice if you do. But they only take up a very small amount of memory.

erikbern:
On #3: This breaks backwards compatibility (that's probably why the unit test fails) so can you figure out a way around adding the "parent" field? I don't see it being used in many places so I'm not sure how it works

erikbern:
This looks pretty cool overall. I really want to keep mmap and avoid adding the "parent" field.

There's also a lot of commits in this branch. Seems like only the first one is relevant for the PR. Can you remove the other commits for now? We can deal with them in a subsequent PR.

I hope we can merge this – would be a great addition.

I'm very curious about what you're using this for. Are you using it at Houzz?

LongbinChen:
hi, Eric,
   thanks for the comments.
   I am building an image search engine whose core ranking function is knn.
I hope to build something like Solr that has the functionality of
 add/delete/update/filtering/sort, and probably with replication if i have
to shard the index.

   I like the mmap which can reduce the memory requirement when serving a
large index. The function "add_item_to_index" would require the serving
machine have big memories which might be a bummer. i will try to make it
work so that even when building index, we don't need a large memory.

  I didn't run the unit test because I don't know how to. can you let me
know how to run the unit test? thanks.


On Sat, Feb 28, 2015 at 9:24 AM, Erik Bernhardsson <notifications@github.com

erikbern:
Ok, sounds great. Maybe I'll play around with your code. I might not understand all the complexity yet.

You should be able to run the unit tests using `nosetests`

erikbern:
I want to pick this up. I might put together a separate PR and borrow a bit from this code. Incremental adds would be a great new feature.

LongbinChen:
i have been looking into using LMDB as the backend storage for the index,
instead of using a simply memory-map array. There are some pros/cons for
using LMDB. Downside is that there would be some memory overheads. However
good side is that LMDB is a well-tested well-maintained memory-map
key-value store, with transaction support. I am afraid if we spend a lot of
time working on the Annoy's own memory map file operations, it is actually
just re-inventing the wheels of LMDB.

Thoughts?

On Sat, Aug 15, 2015 at 3:47 PM, Erik Bernhardsson <notifications@github.com

LongbinChen:
one problem with Annoy we encountered in our system is that when we tried
to build a super large index, the system crashed because building indexing
uses the physical memory, not memory-map file.  Using LMDB could solve the
problem. Of course, we might work with memory array however it would be
much harder to work and the codes would be harder to maintain.

erikbern:
I have been thinking about that case too. I think it's not too hard to fix
actually. Unfortunately mmap is not possible to resize, but you could
imagine splitting the memory in two slots where the first one is mmap and
second is dynamic. Then spill the second one to disk when needed. Or
something like that.

I never heard about LMDB but if it solves the issue then absolutely.

Either way my first step is probably to implement incremental add_item
after build() but before save(). This way I can avoid the memory management
issues for now.

erikbern:
Read some more about LMDB. It seems interesting. I wonder what the overhead would be for int keys. But using non-integer keys has been a requested feature, so maybe it's fine incurring a small speed penalty

wanderer2:
Hello Erik. Do you plan to integrate the proposed incremental update support?

erikbern:
no plans

mostafa8026:
It will be a great feature, if you plan to add incremental build. please take a look again.

PR #50, opened 2015-04-29
Boost dependences is removed

dvenum:
1. I added -std=c++11 flag for std::rand support. Tell me, if it matter.

2. EuclideanIndexTest::test_get_nns_by_vector failed, see below. It's in reverse order, but AngularIndexTest::, passed ok. I don't know, how I can receive [1,0], if it include sort call before return.
annoylib.h:
537     sort(nns_dist.begin(), nns_dist.end());
538     for (size_t i = 0; i < nns_dist.size() && result->size() < n; i++) {
539       result->push_back(nns_dist[i].second);
540     }

FAIL: test_get_nns_by_vector (annoy_test.EuclideanIndexTest)

  File "/home/dvenum/devl/annoy/test/annoy_test.py", line 112, in test_get_nns_by_vector
    self.assertEqual(i.get_nns_by_vector([3,3], 2), [1, 0])
AssertionError: Lists differ: [0, 1] != [1, 0]
First differing element 0:
0
1
<div>
-  [0, 1] <br/>
+  [1, 0] <br/>
<br/>
3. If I right, valgrin have clear output, only usually python message. It seems to true, because a small new memory management was delegated to python.
</div>

### DIFF
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -41,9 +41,9 @@ public:
 
   PyObject* get_nns_by_vector_py(PyObject* v, size_t n) {
     vector<T> w(this->_f);
-    for (int z = 0; z < PyList_Size(v); z++) {
+    for (int z = 0; z < PyList_Size(v) && z < this->_f; z++) {
         PyObject *pf = PyList_GetItem(v,z);
-        w.push_back(PyFloat_AsDouble(pf));
+        w[z] = PyFloat_AsDouble(pf);
     }
     vector<S> result;
     this->get_nns_by_vector(&w[0], n, &result);
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -439,44 +439,44 @@ static PyMethodDef AnnoyMethods[] = {
 
 static PyTypeObject PyAnnoyType = {
   PyObject_HEAD_INIT(NULL)
-  0,						 /*ob_size*/
-  "annoy.Annoy",			 /*tp_name*/
-  sizeof(py_annoy),			 /*tp_basicsize*/
-  0,						 /*tp_itemsize*/
+  0,                      /*ob_size*/
+  "annoy.Annoy",          /*tp_name*/
+  sizeof(py_annoy),       /*tp_basicsize*/
+  0,                      /*tp_itemsize*/
   (destructor)py_an_dealloc, /*tp_dealloc*/
-  0,						 /*tp_print*/
-  0,						 /*tp_getattr*/
-  0,						 /*tp_setattr*/
-  0,						 /*tp_compare*/
-  0,						 /*tp_repr*/
-  0,						 /*tp_as_number*/
-  0,						 /*tp_as_sequence*/
-  0,						 /*tp_as_mapping*/
-  0,						 /*tp_hash */
-  0,						 /*tp_call*/
-  0,						 /*tp_str*/
-  0,						 /*tp_getattro*/
-  0,						 /*tp_setattro*/
-  0,						 /*tp_as_buffer*/
+  0,                      /*tp_print*/
+  0,                      /*tp_getattr*/
+  0,                      /*tp_setattr*/
+  0,                      /*tp_compare*/
+  0,                      /*tp_repr*/
+  0,                      /*tp_as_number*/
+  0,                      /*tp_as_sequence*/
+  0,                      /*tp_as_mapping*/
+  0,                      /*tp_hash */
+  0,                      /*tp_call*/
+  0,                      /*tp_str*/
+  0,                      /*tp_getattro*/
+  0,                      /*tp_setattro*/
+  0,                      /*tp_as_buffer*/
   Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE, /*tp_flags*/
-  "annoy objects",			 /* tp_doc */
-  0,						 /* tp_traverse */
-  0,						 /* tp_clear */
-  0,						 /* tp_richcompare */
-  0,						 /* tp_weaklistoffset */
-  0,						 /* tp_iter */
-  0,						 /* tp_iternext */
-  AnnoyMethods,				 /* tp_methods */
-  py_annoy_members,			 /* tp_members */
-  0,						 /* tp_getset */
-  0,						 /* tp_base */
-  0,						 /* tp_dict */
-  0,						 /* tp_descr_get */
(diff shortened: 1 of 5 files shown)


erikbern:
Looks good.

The sort on line 537 sorts by the first element of each pair (the distance) which is smaller for element 1 than for element 0.

Element 1 should have squared distance 2 and element 0 should have squared distance 3.

dvenum:
It seems to my bad. nns_dist contain:
8 0
13 1

But for angular test: 
2 0
2 1
2 2

erikbern:
Look at https://github.com/spotify/annoy/blob/master/test/annoy_test.py#L105

`nn_dist` should contain
1 1
2 0

(sorry I incorrectly said 2 and 3 in my previous comment, that was not right)

src/annoymodule.cc
```diff
@@ -14,87 +14,490 @@
...
+    ai = static_cast<AI*>(self->ptr);
+    res = ai->load_py(filename);
+    break;
+  case 'e':
+    ei = static_cast<EI*>(self->ptr);
+    res = ei->load_py(filename);
+    break;
+  }
+  if (!res) {
+    PyErr_SetString(PyExc_IOError,"load error");
```
a1k0n: It should be safe to use PyErr_SetFromErrno(PyExc_IOError); here, as errno should be preserved from the load_py call. I think.

src/annoymodule.cc
```diff
@@ -14,87 +14,490 @@
...
+  0,						 /* tp_getset */
+  0,						 /* tp_base */
+  0,						 /* tp_dict */
+  0,						 /* tp_descr_get */
+  0,						 /* tp_descr_set */
+  0,						 /* tp_dictoffset */
+  (initproc)py_an_init,		 /* tp_init */
+  0,						 /* tp_alloc */
+  py_an_new,				 /* tp_new */
+};
```
a1k0n: would you mind expanding the tabs to spaces above here?

src/annoymodule.cc
```diff
@@ -14,87 +14,490 @@
...
+    ai = static_cast<AI*>(self->ptr);
+    res = ai->save_py(filename);
+    break;
+  case 'e':
+    ei = static_cast<EI*>(self->ptr);
+    res = ei->save_py(filename);
+    break;
+  }
+  if (!res) {
+    PyErr_SetString(PyExc_IOError,"load error");
```
a1k0n: `PyErr_SetFromErrno(PyExc_IOError);` here too, would be more informative.

a1k0n:
If AI and EI were subclasses with virtual member functions, you could save a whole lot of dispatch code here.

erikbern:
As @a1k0n pointed out making AI and EI subclasses with virtual functions seems like the easiest way to avoid the switching and static casting.

Another way is to move the whole switching to compile time and compile two different python extensions. But that's a bit more complex and the performance gains are pretty irrelevant

dvenum:
As @erikbern pointed me, it will similar to:
// annoy python object
template<typename S, typename T, typename Distance>
typedef struct {
    PyObject_HEAD
    int f;
    char ch_metric;
    AnnoyIndexPython<S, T, Distance >* ptr;
} py_annoy;

I was a little confusing with templates exposing. It was first working code with rewriting purpose.

src/annoymodule.cc
dvenum: It's done, sorry.

dvenum:
Critical bugfix in c02bb89. Input vector was always zero.
test_get_nns_by_vector passed.

erikbern:
I don't understand why it solves the problem but great that it works now!

dvenum:
push_back was used me for vector object when it convering from input python list. New items was added to end, but it using by index.
```c
    vector<T> w(this->_f); <<-- start length no zero
    for (int z = 0; z < PyList_Size(v) && z < this->_f; z++) {   <<-- check border also
        PyObject *pf = PyList_GetItem(v,z);
        w[z] = PyFloat_AsDouble(pf);             <<-- correct way
    }
```

erikbern:
oh ok, makes sense! of course. thanks for the explanation!

dvenum:
I have same problem with templates.

As below, object py_annoy should be allocated when I don't know, what is object required. No metrics received yet. ptr should be a pointer to an universal base class.
```c
// annoy python object
template<typename S, typename T, typename Distance>
struct py_annoy_ { 
  PyObject_HEAD
  int f;
  char ch_metric;
  void* ptr;
  AnnoyIndexPython<S, T, Distance >* ptr;
};

static PyObject *
py_an_new(PyTypeObject *type, PyObject *args, PyObject *kwds) {
  py_annoy *self;

  self = (py_annoy *)type->tp_alloc(type, 0);
  if (self != NULL) {
    self->f = 0;
    self->ch_metric = 0;
    self->ptr = NULL;
  }

  return (PyObject *)self;
}
```

Boost determine class template, which may wrap C++ class and expose him. It's a bit complex way for this task.
I think about some tricks in C-style. It'll ok when we will have few wrapper functions if no code will duplicated.

erikbern:
Ok, that's fine, I can take a look at it later.

Can you remove the "-std=c++11" flag though? Travis-CI doesn't like it for some reason

dvenum:
<random> lib exists only for c++11. I can include another library as part of sources if nothing to more will be.

May you try something, similar to it in .travis.yml?
```bash
install:
  - sudo apt-get install -qq gcc-4.8 g++-4.8 
# We want to compile with g++ 4.8 when rather than the default g++
  - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90
```

erikbern:
Which lib exists only for c++11? I think `#include <random>` should work for older versions too

dvenum:
As you can see on the head, only C++11 marker exists: 
http://www.cplusplus.com/reference/random/

Old C++ have only rand() function from stdlib.

erikbern:
Ok maybe we just have to include something like this: http://phoxis.org/2013/05/04/generating-random-numbers-from-normal-distribution-in-c/

dvenum:
If using rand() call is correct, I can rewrite it, no problem. Or I can include a bit more complex library.
Travis-CI uses gcc-4.6, that very old.

erikbern:
I can fix. Let's merge this now!

erikbern:
It was pretty easy: https://github.com/spotify/annoy/pull/52

PR #52, opened 2015-04-30
remove dependency on <random> and c++11

erikbern:


dvenum:
;). Nice. 
rand() no good generator. I glad that it's enough for task.

Leave "depends=['src/annoylib.h']," please. It's good, when you can rebuild after change in header.

src/annoylib.h
```diff
@@ -62,22 +61,36 @@ template<typename T>
...
-    return _var_nor(generator);
+    // Taken from http://phoxis.org/2013/05/04/generating-random-numbers-from-normal-distribution-in-c/
+    _has_X2 = !_has_X2;
+    if (_has_X2)
+      return _X2;
+
+    T W, U1, U2;
+    do {
+      U1 = -1 + ((T)rand() / RAND_MAX) * 2;
+      U2 = -1 + ((T)rand() / RAND_MAX) * 2;
```
a1k0n: fwiw drand48() is a better random number generator and should be widely compatible but... this is fine.

erikbern: it's just for random splits anyway, should be fine

actually the reason i use normal distribution is just that it's a pretty easy way to generate random unit vectors. maybe there's some even simpler way

PR #57, opened 2015-05-02
python3 support

dvenum:
All tests passed.
Don't worry about new library name annoylib.cpython-34m.so (for python34), it's imported as well.
http://legacy.python.org/dev/peps/pep-3149/

sorry, annoylib.h changes was included here.


erikbern:
nice! i will add python 3 to travis as well

PR #59, opened 2015-05-02
allow alternate uniform RNG

eddelbuettel:
This PR allows RcppAnnoy to bring in R's uniform RNG with next-to-no effort via

```{.c++}
// define R's unif_rand() as the uniform rand generator for Annoy
#define __UNIFORM_RAND_OVERRIDE__ static_cast<long int>(unif_rand() * RAND_MAX)
```

which is a good thing.


erikbern:
Looks good although minor complaint is I would have preferred if you had squashed your three commits into one

PR #60, opened 2015-05-03
correct placement of curly brace

eddelbuettel:
The change is line 58 is cosmetic; the change in line 62 is required as the real closing brace is line 64 is outside the `#ifdef` / `#endif` block.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -55,11 +55,9 @@
 // which is assumed to be uniformly distributed over over 0 .. RAND_MAX
 inline long int randomDraw() {
 #ifndef __UNIFORM_RAND_OVERRIDE__
-  // by default we use random(); if for some reason you do not have random() try rand()
   return rand() * RAND_MAX;
 #else
   return __UNIFORM_RAND_OVERRIDE__;
-}
 #endif
 }
 


eddelbuettel:
Thanks.  Do you think you'll call this 1.2.2?

erikbern:
What's the impact of the fix? If it solves a real issue I can upload a new version to pypi

erikbern:
looks like it solves a compilation issue if the rng is overridden?

eddelbuettel:
I don't understand how I didn't catch that yesterday as I kept the `annoylib.h` in sync -- but yes, when the `#define` I use is set, we end up with two `}` and hence in tears.

It would be cleaner to roll up a new tarball.  And heck, that is what the third digit is for :)

erikbern:
there you go

eddelbuettel:
:+1: 

Will roll up RcppAnnoy 0.0.6 once I see 1.2.2 -- so far I only see a change to the `README.rst`

erikbern:
1.2.2 is out!

erikbern:
actually forgot to push setup.py to github, done now

eddelbuettel:
See, I was watching https://github.com/spotify/annoy/blob/master/setup.py#L34 and it had not moved :)

Issue #64, opened 2015-05-22
idea for faster/better tree structure


erikbern:
What if at every split we just sample two items and compute the hyperplane that separates them?

For angular distance you would normalize the two items first. This would also generalize to arbitrary pseudometrics nicely: Manhattan, KL-divergence, etc.

For highdimensional distributions where there's structure in the data, this is more likely to sample from that structure. It will also be much faster in general since we are guaranteed to split the point set in two subsets

a1k0n:
Yeah, that makes a lot of sense.

I mean, you could also take a random direction and then compute the dot product on all points in the leaf and find the median, and do that over and over, but for angular distance you'd have to introduce a bias coefficient to make that work and I don't know if that destroys the cosine distance guarantees you get with signed random projections. It would be better for the euclidean one though.

This is a really neat idea though.

erikbern:
I'll think of a way to test how well it works.

erikbern:
See #65 – results are *very* good

stefansavev:
Hi Eric,

First congratulations on the excellent results with the new method. Just wanted to point out a paper by Sanjoy Dasgupta where he considers a related rule. May be the paper will useful.
http://cseweb.ucsd.edu/~dasgupta/papers/rptree-stoc.pdf (Random projection trees and low dimensional manifolds)
You can look for the text: "choose a random unit direction v ∈ R; pick any x ∈ S; let y ∈ S be the farthest point from it"
No idea how it relates performance-wise to your rule.

Cheers,

Stefan

erikbern:
Thanks a lot – looks pretty interesting! Here's the relevant snippet (I think)

![image](https://cloud.githubusercontent.com/assets/1027979/8862091/189438f6-315c-11e5-849c-90340ccc8638.png)

I'll implement this at some point and benchmark!

a1k0n:
That method still picks a uniformly random direction; the points x and y are only used to find the split plane along the direction, which annoy doesn't really support -- the split plane is always at 0.

erikbern:
@a1k0n – that's only for cosine. For Euclidean, you can pick any split plane

a1k0n:
@erikbern of course. Yes, this might make sense for the Euclidean method. I don't follow the argument in the paper yet but it's very interesting. Thanks @stefansavev

erikbern:
[...]

PR #65, opened 2015-05-25
faster/better tree building algorithm (see #64)

erikbern:
I haven't validated this with any real dataset, need to do that

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -75,41 +75,25 @@ struct Randomness {
   T _X1, _X2;
   bool _has_X2;
 
-  inline T gaussian() {
-    // Taken from http://phoxis.org/2013/05/04/generating-random-numbers-from-normal-distribution-in-c/
-    _has_X2 = !_has_X2;
-    if (_has_X2)
-      return _X2;
-
-    T W, U1, U2;
-    do {
-      U1 = -1 + ((T) randomDraw() / RAND_MAX) * 2;
-      U2 = -1 + ((T) randomDraw() / RAND_MAX) * 2;
-      W = U1 * U1 + U2 * U2;
-    }
-    while (W >= 1 || W == 0);
-
-    T mult = sqrt((-2 * log(W)) / W);
-    _X1 = U1 * mult;
-    _X2 = U2 * mult;
-
-    return _X1;
-  }
-
   inline int flip() {
     return randomDraw() % 2;
   }
-  inline T uniform(T min, T max) {
-    return ((T) randomDraw() / RAND_MAX) * (max - min) + min;
+  inline int index(int n) {
+    return randomDraw() % n;
   }
 };
 
 template<typename T>
-inline void normalize(T* v, int f) {
+inline T get_norm(T* v, int f) {
   T sq_norm = 0;
   for (int z = 0; z < f; z++)
     sq_norm += v[z] * v[z];
-  T norm = sqrt(sq_norm);
+  return sqrt(sq_norm);
+}
+
+template<typename T>
+inline void normalize(T* v, int f) {
+  T norm = get_norm(v, f);
   for (int z = 0; z < f; z++)
     v[z] /= norm;
 }
@@ -164,8 +148,14 @@ struct Angular {
       return random->flip();
   }
   static inline void create_split(const vector<node*>& nodes, int f, Randomness<T>* random, node* n) {
+    // Sample two random points from the set of nodes
+    // Calculate the hyperplane equidistant from them
+    int i = random->index(nodes.size());
+    int j = random->index(nodes.size());
+    T i_norm = get_norm(nodes[i]->v, f);
+    T j_norm = get_norm(nodes[j]->v, f);
     for (int z = 0; z < f; z++)
-      n->v[z] = random->gaussian();
+      n->v[z] = nodes[i]->v[z] / i_norm - nodes[j]->v[z] / j_norm;
     normalize(n->v, f);
   }
 };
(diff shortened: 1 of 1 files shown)


erikbern:
I want to run this on a real dataset to see if it helps.

I'm also a bit concern what happens in degenerate cases such as `i==j` but I think it will work out fine

erikbern:
Before 
```
Ran 21 tests in 115.291s
```

After
```
Ran 21 tests in 23.230s
```

erikbern:
See accuracy test that I just added: https://github.com/spotify/annoy/commit/f5c1f58df11cc156aad8e6d73f809c8bccdae7c7

The improvements are pretty spectacular:

Before
```
             angular   25 accuracy: 34.64%
             angular   50 accuracy: 20.44%
             angular  100 accuracy: 12.90%
             angular  200 accuracy:  7.99%

           euclidean   25 accuracy: 33.71%
           euclidean   50 accuracy: 21.04%
           euclidean  100 accuracy: 13.37%
           euclidean  200 accuracy:  9.31%
```

After
```
             angular   25 accuracy: 46.80%
             angular   50 accuracy: 31.00%
             angular  100 accuracy: 24.50%
             angular  200 accuracy: 15.18%

           euclidean   25 accuracy: 47.34%
           euclidean   50 accuracy: 33.04%
           euclidean  100 accuracy: 25.10%
           euclidean  200 accuracy: 17.99%
```

The runtime of the tests also dropped from about 6000s to 1500s meaning a speedup of 4x.

Pretty cool for a code review that's mostly red!

src/annoylib.h
```diff
@@ -198,35 +188,14 @@ struct Euclidean {
...
-          min = dot;
-      }
-      if (max - min > random->uniform(0, max_proj)) {
-        max_proj = max - min;
-        memcpy(n->v, v, sizeof(T) * f);
-        n->a = -random->uniform(min, max); // Take a random split along this axis
-      }
+    // Same as Angular version, but no normalization and has to compute the offset too
+    int i = random->index(nodes.size());
+    int j = random->index(nodes.size());
```
a1k0n: what if i == j?

src/annoylib.h
```diff
@@ -164,8 +148,14 @@ struct Angular {
       return random->flip();
   }
   static inline void create_split(const vector<node*>& nodes, int f, Randomness<T>* random, node* n) {
+    // Sample two random points from the set of nodes
+    // Calculate the hyperplane equidistant from them
+    int i = random->index(nodes.size());
+    int j = random->index(nodes.size());
```
a1k0n: what if i == j?

you'll have to sample nodes.size()-1 for j, and if(j >= i) j++;

a1k0n:
That is super awesome, though.

src/annoylib.h
erikbern: There is an outer loop that will repeat the sampling up to 10 times anyway, so it shouldn't be a problem.

But sure maybe I should fix it

a1k0n: Oh, right, that's true.

You could also enforce slightly more rigorous splits, i.e. at least 80%/20%, yielding better accuracy at the cost of taking longer to build the tree.

erikbern: I fixed it later.

Yes you can do something like that too. I wonder if there are some cheap tricks in practice where you can compute statistics on subsamples, eg. sample 10 random points and pick the pair of points that gives the most even split. Sounds a bit more messy to implement though!

MrChrisJohnson: Not sure the probability of choosing a poor hyperplane by sampling only 2 points, but if you sample 10 then it's pretty easy to find a hyperplane that splits the 10 evenly. You sample a random hyperplane, project all 10 points to the randomly chosen hyperplane, then choose a hyperplane perpendicular to this one that passes through the original splitting the 10 projected points evenly. Make sense? Easier to draw on paper then to describe here.

I'd be careful using a fully randomized Las Vegas algorithm (reject if < 80/20 split) since it costs O(n) to verify (where n is the number of points left to split in the subtree). If you end up in a loop of rejects then this can be quite costly in the early splits, but sampling more points on the initial hyperplane choice is much less costly, and if they are chosen uniformly then more or less likely to represent the full distribution.

erikbern: Yes it makes sense. I have no idea what's the implications on runtime of any of these ideas though – would have to implement it and try. The beauty of new method is how trivial it is (only a few lines).

One pitfall if you try to be "too smart" is also that the splits end up being too nonrandom. You still need to have a fair amount of randomness to decorrelate the trees. There might be some smart way to achieve the same thing but I have no idea what it would be

a1k0n: Chris: that makes sense for the euclidean case but it's less clear how it works in the angular case. But you're right, the las vegas algorithm would be much, much more costly unless you subsample first. Hmm.

MrChrisJohnson: Hmm, maybe I am misunderstanding, but shouldn't it always end up with an even split since it's a hyperplane and the similarity metric won't define what side of the hyperplane a point lies? I think what you are worried about is that it could unfairly split similar points because they could potentially seem like they are far from each when you project onto the randomly chosen plane. ie: if the angle between 2 points was small but the magnitude was large. In that case, as long as you normalize ahead of time it should be a more fair split since then all points will lie on the unit sphere. Might need to think about this some more. 

Good point about the simplicity and avoiding any determinism in the tree splits. Not sure on the theoretical implications of smarter splits, but seems like there will still be plenty of randomness from the initial hyperplane choice as well as the samples.

erikbern: It's not always an even split – you could end up having 1 point on one side and n-1 points on the other side. Or what did you mean?

Yes you need to be careful about not consistently separating the same points that are close to each other. That's why you need to build an index with multiple trees. Any given tree might separate two close points, but it's unlikely they are separated in all trees

MrChrisJohnson: I just meant for the proposed sampling method above it will always split the 10 sampled points evenly regardless of similarity metric (obviously not the full set of points if you are only sampling 10). 

Yep, completely agree. Not fully sure on the theoretical implications of the sampling method but seems like a nice balance of being smarter about the hyperplane split (splitting k points instead of just 2) but also maintaining nondeterminism so as to avoid trees replicating splits.

erikbern: Btw I think an easy generalization from 2 to k would be: sample k (where k is even) points, compute the principal component, split on the median.

a1k0n: I thought of that too but how do you get the principal component? You either need to store a d x d matrix or do power iterations I guess?

erikbern: was thinking power iterations. i think you can do it in a few lines but i don't remember the math.

a1k0n: you'd start with an initial guess, maybe ones(d), and repeatedly multiply it by the covariance matrix from summing up the outer products of all the vectors, which is the same as adding up the dot product of the guess with the outer product of each vector with itself... here: http://www.texpaste.com/n/7ao61x20

MrChrisJohnson: Seems doable, you could potentially even calculate several splits at once (take the top 3 principle components). One thing to keep in mind with PCA though is that it's susceptible to outliers so you aren't really guaranteed an even split, especially if your sample isn't very gaussian-like. In that case, to maximize the variance it may find a split that overemphasizes 1 or 2 outliers in your sampled set.

erikbern: that makes sense. you could even subsample the indexes within each power iteration. so every iteration you just multiply with a bunch of randomly sampled vectors.

i can try this later this week, should be a few lines of code

MrChrisJohnson: Still, I'm unsure why you'd prefer PCA to the simple O(k) method above. Less costly, simple to implement, and will always guarantee an even split.

a1k0n: But if you subsample different points each time you either converge to the optimum split plane, which means your different random trees wouldn't be so random anymore, or you don't converge at all... I think the benefit from the random forest comes from sampling a specific number of points, and as Chris says, you might as well just keep a small sample, and maybe that sample being 2 points is good enough anyway.

erikbern: @MrChrisJohnson not sure if I understand the O(k) algo you're describing

@a1k0n yes as i mentioned the reason you use different trees is to make sure you don't consistently separate the same two close points

but so far annoy has erred too much on the side of uncorrelated trees and too little on using the  distribution of the points. i think you could probably go a little further leveraging the distribution of the points

i'll prob just keep the 2-point solution for now. should be enough

MrChrisJohnson: @erikbern The O(k) method is to sample k points as well as a random hyperplane h. Then, project each of the k points to the random hyperplane (in 2 dimensions this is just a bunch of points on a line). It should now be trivial to find a second hyperplane h' perpendicular to h that splits the projected points and will in turn split the points in the original space. 

I tried to draw a picture of this. Not sure if it came through.

image

erikbern: Ok I get what you are saying. Makes sense.

I think you still want to pick the first vector not uniformly randomly but somehow skewed to the distribution of the data set. If you have a very high dimensional dataset that is basically some low dimensional hyperplane H plus a bunch of noise then the new 2-point sampling algorithm will generate splits approximately perpendicular to H which seems beneficial

PR #75, opened 2015-06-08
Make add_item() not verbose

atqamar:
Do not return and print 0's when add_item() is called.

Change-Id: Ieb8e3ee43ab3bcaff908c71ba99624ced86e49c4

### DIFF
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -205,7 +205,7 @@ py_an_add_item(py_annoy *self, PyObject *args) {
   }
   self->ptr->add_item(item, &w[0]);
 
-  return PyInt_FromLong(0);
+  Py_RETURN_NONE;
 }
 
 


src/annoymodule.cc
```diff
@@ -205,7 +205,6 @@ py_an_add_item(py_annoy *self, PyObject *args) {
   }
   self->ptr->add_item(item, &w[0]);
 
-  return PyInt_FromLong(0);
```
erikbern: I think this could cause nasty things – you have a function that should return a PyObject* but without a "return" statement it will just read some random memory off the stack and treat as a PyObject*

In Python returning None is the same as no return statement so I think you can do `return Py_None;` here and everything should work

erikbern: yeah you got a bunch of segfaults... but try what I said and it should work

atqamar: Ahh yes, it's not a void typed function. Thanks.

It's working now.

PR #83, opened 2015-06-26
Slightly faster searches

erikbern:
Instead of just putting the margin into the heap, we put the minimum margin seen across all nodes in the path from the root.

On my computer it takes the test time from 70s -> 62s which is a 13% improvement. Not huge but not bad for such a small fix


erikbern:
Actually running ann-benchmarks shows pretty substantial improvements: +50-100% in most cases

image

Interestingly it's actually on par or faster than kgraph. The only faster algorithm is SW-graph from nmslib

image

Note that some of the increase in performance might be due to #82 as well

PR #84, opened 2015-07-05
Filter by number of trees

erikbern:
Computing distances is the biggest bottleneck in the searching. It turns out that even before computing distances, we have a decent proxy for it: the number of trees that any item was found in during search.

This PR change the logic so we only compute distances for items that occurred in at least T trees during the search. T is determined in such a way so that it will compute ``search_m = sqrt(n * search_k)`` distances. This seems reasonable although there's no derivation of it.

Benchmarking on my laptop this seems more than 2x faster at no impact on precision.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -503,6 +503,8 @@ protected:
     if (search_k == -1)
       search_k = n * _roots.size(); // slightly arbitrary default value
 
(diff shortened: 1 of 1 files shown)


erikbern:
Seems like this introduced a regression from 0.98 to 0.97 for the `test_precision` tests... it's probably fine

erikbern:
This doesn't look good in benchmarks – I will trouble shoot it

image

PR #89, opened 2015-07-25
Support non-list types

erikbern:
By converting to a list type, it's now possible to insert numpy arrays etc. This would previously lead to a segfault.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - "3.4"
 
 install:
+  - pip install numpy
   - pip install .
 
 script: python setup.py nosetests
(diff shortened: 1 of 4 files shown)


piskvorky:
Nice. Calling `list(numpy array)` doesn't change the dtype of its elements though. It's still a list of `numpy.float64` or `float32` or whatever. 

I forgot how Annoy handled such cases, but isn't there a potential for segfault as well?

IIRC I had to always manually re-cast Annoy inputs to `float`, in my benchmarks.

erikbern:
i can add some tests for other dtypes!

erikbern:
but i don't think it should matter. if you cast it to a list, it will probably cast all the underlying elements too right? there's no unboxed float32 in Python i think

erikbern:
Using various numpy types in the test now

piskvorky:
No, it won't. It is a common grievance, because you end up with numpy types inside plain Python structures.

Sometimes you don't notice until you try to serialize them and things blow up, such as when sending the objects through an HTTP interface. For example, json cannot serialize numpy types, and will throw an exception that will leave you scratching you head, unless you've seen it before :)

```python
>>> a = numpy.arange(10).astype(numpy.float32)
>>> l = list(a)
>>> json.dumps(l)
TypeError: 0.0 is not JSON serializable
```

erikbern:
That's annoying. Seems like it's not an issue here though

piskvorky:
Yeah, I saw you added unit tests for that, good idea.

There must be some NumPy / Python C API conversion magic that takes care of this implicitly... though I don't recall where exactly :)

PR #95, opened 2015-08-12
Adds KISS random number generators

Jochen0x90h:
Now AnnoyIndex has an additional template parameter to pass it a random number generator. Custom random number generators may have their own internal state. There are three generators: rand() based, KISS32 and KISS64.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -23,6 +23,8 @@
 #include <stdlib.h>
 #include <sys/types.h>
 #include <fcntl.h>
+#include <stddef.h>
+#include <stdint.h>
 
 #ifdef __MINGW32__
 #include "mman.h"
@@ -46,20 +48,11 @@
   #define showUpdate(...) { __ERROR_PRINTER_OVERRIDE__( __VA_ARGS__ ); }
 #endif
 
-#ifndef NO_PACKED_STRUCTS
-#define PACKED_STRUCTS_EXTRA __attribute__((__packed__))
-// TODO: this is turned on by default, but may not work for all architectures! Need to investigate.
+#ifndef ANNOY_NODE_ATTRIBUTE
+  #define ANNOY_NODE_ATTRIBUTE __attribute__((__packed__))
+  // TODO: this is turned on by default, but may not work for all architectures! Need to investigate.
 #endif
 
-// This allows others to supply their own random number generator function
-// which is assumed to be uniformly distributed over over 0 .. RAND_MAX
-inline long int randomDraw() {
-#ifndef __UNIFORM_RAND_OVERRIDE__
-  return rand();
-#else
-  return __UNIFORM_RAND_OVERRIDE__;
-#endif
-}
 
 using std::vector;
 using std::string;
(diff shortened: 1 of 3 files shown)


erikbern:
looks good but can you put the random number generates in a separate header file?

erikbern:
also would be good to understand why this is useful

src/annoylib.h
```diff
@@ -457,7 +513,14 @@ template<typename S, typename T, typename Distance>
           children.push_back(n);
       }
 
-      Distance::create_split(children, _f, &_random, m);
+      {
+        // Sample two random points from the set of nodes so that i != j
```
erikbern: i'm not sure about this refactoring – can you explain why you moved the sampling to here instead?

src/annoylib.h
```diff
@@ -67,21 +58,89 @@ using std::pair;
...
+
+    return x + y + z;
   }
-  inline int index(int n) {
-    return randomDraw() % n;
+  inline int flip() {
+    // Draw random 0 or 1
+    return kiss() & 1;
   }
-  inline void index2(int n, int* i, int* j) {
```
erikbern: why did you remove this method?

erikbern:
btw separately i want to templatize in terms of `S` as well but i can do that some other time

src/annoylib.h
Jochen0x90h: It makes it easier to implement a random number generator. The sampling of two points seems to be done only at this place so why not have the code here?

Jochen0x90h:
Yes, having the extra random number generators in an extra header is a good idea. I personally don't trust random functions from standard libraries. But it could be possible to remove Kiss altogether again and keep only the default rand() based implementation. Then I can add Kiss myself. Templating for S may not be necessary if you pick a random number generator matching for S (e.g. Kiss32 for S = int32_t). But for convenience there could be a KissRandom<S> that selects 32 or 64 bit according to S using partial specialization.

src/annoylib.h
erikbern: i think that's a leaky abstraction – what if at some point later another distance function has a very different way to create a split plane?

erikbern:
ok yeah that makes sense. yeah maybe the templatizing is the main thing you need at this point. but if `S` is 64 bit it would be useful to have a better RNG.

Actually even if S is smaller than `2^32` it probably makes sense to always use the 64 bit RNG since when you are doing modulo it will create a distribution that's not uniform (except if number of items is a power of two)

Jochen0x90h:
If the number of data points is much smaller than 2^32 then 32 bit RNG is ok too. By the way the standard guarantees only that RAND_MAX is at least 32767 (15 bit RNG). I now moved the RNGs into a separate file and reverted create_split, but implemented the picking of the two points directly without the original index2 function. I hope that's ok.

erikbern:
can you squash the first three commits?

erikbern:
Can you fix compiler errors (check Travis) and rebase your commits? Thanks!

Jochen0x90h:
Now I hope it can be merged

erikbern:
LGTM!

erikbern:
interestingly the accuracy went up marginally in the unit test. probably just a fluke

Jochen0x90h:
For me it also marginally became better but no surprise as I don't believe in system RNGs ;)

erikbern:
do you think we should default to the 64 bit RNG? if you have say 1B nodes then you will get a really skewed distribution with a 32 bit RNG

Jochen0x90h:
Yes, maybe it's better to use 64 bit. Shouldn't hurt on today's 64 bit machines if it looks good in the unit test

Jochen0x90h:
Now it's ready I think

PR #101, opened 2015-09-09
incomplete change

LongbinChen:
hi, Eric, please feel free to take whatever you need.


erikbern:
This looks great!

Did you consider using a plain old struct instead of protobuf? It's probably faster.

It might be worth just starting over with a new repository and calling it annoy2 or annul or something :)

LongbinChen:
yes. i debated a lot.

protobuf would have very nice interface for whoever want to add new
indexing field into the system later; also it saves the work for
seriazation and deserialzation.

the downside is of course the overhead. however i think protobuf is highly
optimized so it is probably fine.

agree we can start another repo.

LongbinChen:
maybe we can throw in a flask server to make it a full blown search engine
based on distance.

erikbern:
Sure SGTM although I think in practice people might be fine just using the Python bindings and build a simple server around it

erikbern:
If you want to move this to houzz/xyz, I'm happy to help. Was planning to do something under erikbern/ but I've been postponing it so long that it's better if someone else takes charge and I'm happy to contribute to it

LongbinChen:
great. will do it soon.

LongbinChen:
up now.

https://github.com/Houzz/annoy2

piskvorky:
Just wondering... what does the move entail?

To be honest `erikbern/annoy` sounds a little safer to me, but I may not understand all the implications here.

LongbinChen:
it is not quite about being safer or more danger. The codes are open so
there is not really much difference.  The houzz/annoy2 would be the place
for developing lmdb backend. if lmdb backend works, we can either merge it
back or to transfer to somewhere else.

erikbern:
SGTM

The LMDB backend will have a different interface and indexes will not be binary compatible so I think it's the right way to go.

PR #104, opened 2015-09-23
Normalize Euclidean splits

erikbern:
Seems marginally faster. For one test I ran, accuracy went from 89.10% to 89.89% and time spent from 18.211 to 17.862. So a couple of percent better.


a1k0n:
Huh. Shouldn't make much difference, but OK!

erikbern:
It will affect the priority queue during searching. Nodes will now be
explored in order by distance to split.

a1k0n:
Ah, of course. LGTM.

PR #105, opened 2015-10-12
Use k-means to construct hyperplanes

erikbern:
This seems like a pretty huge improvement according to some benchmarks I ran.

```
Euclidean, f=100, GloVe: 18.211s -> 16.573s, 89.10% -> 95.98%
Angular,   f=100, GloVe: 38.947s -> 32.799s, 89.07% -> 96.74%
```

Only drawback is tree building is marginally slower

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -88,6 +88,42 @@ inline void normalize(T* v, int f) {
     v[z] /= norm;
 }
 
+template<typename T, typename Random, typename Distance, typename Node>
+inline void two_means(const vector<Node*>& nodes, int f, Random& random, bool cosine, T* iv, T* jv) {
+  /*
+    This algorithm is a huge heuristic. Empirically it works really well, but I
+    can't motivate it well. The basic idea is to keep two centroids and assign
+    points to either one of them. We weight each centroid by the number of points
+    assigned to it, so to balance it. 
+  */
+  static int iteration_steps = 200;
+  size_t count = nodes.size();
+
+  size_t i = random.index(count);
+  size_t j = random.index(count-1);
+  j += (j >= i); // ensure that i != j
+  std::copy(&nodes[i]->v[0], &nodes[i]->v[f], &iv[0]);
+  std::copy(&nodes[j]->v[0], &nodes[j]->v[f], &jv[0]);
+  if (cosine) { normalize(&iv[0], f); normalize(&jv[0], f); }
+
+  int ic = 1, jc = 1;
+  for (int l = 0; l < iteration_steps; l++) {
+    size_t k = random.index(count);
+    T di = ic * Distance::distance(&iv[0], nodes[k]->v, f),
+      dj = jc * Distance::distance(&jv[0], nodes[k]->v, f);
+    T norm = cosine ? get_norm(nodes[k]->v, f) : 1.0;
+    if (di < dj) {
+      for (int z = 0; z < f; z++)
+	iv[z] = (iv[z] * ic + nodes[k]->v[z] / norm) / (ic + 1);
+      ic++;
+    } else if (dj < di) {
+      for (int z = 0; z < f; z++)
+	jv[z] = (jv[z] * jc + nodes[k]->v[z] / norm) / (jc + 1);
+      jc++;
+    }
+  }
+}
+
 
 struct Angular {
   template<typename S, typename T>
@@ -142,19 +178,10 @@ struct Angular {
   }
   template<typename S, typename T, typename Random>
   static inline void create_split(const vector<Node<S, T>*>& nodes, int f, Random& random, Node<S, T>* n) {
-    // Sample two random points from the set of nodes
-    // Calculate the hyperplane equidistant from them
-    size_t count = nodes.size();
-    size_t i = random.index(count);
-    size_t j = random.index(count-1);
-    j += (j >= i); // ensure that i != j
-    T* iv = nodes[i]->v;
-    T* jv = nodes[j]->v;
-    T i_norm = get_norm(iv, f);
-    T j_norm = get_norm(jv, f);
+    static vector<T> best_iv(f, 0), best_jv(f, 0);
+    two_means<T, Random, Angular, Node<S, T> >(nodes, f, random, true, &best_iv[0], &best_jv[0]);
     for (int z = 0; z < f; z++)
-      n->v[z] = iv[z] / i_norm - jv[z] / j_norm;
-    normalize(n->v, f);
+      n->v[z] = best_iv[z] - best_jv[z];
   }
   template<typename T>
   static inline T normalized_distance(T distance) {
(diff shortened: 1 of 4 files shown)


src/annoylib.h
```diff
@@ -88,6 +88,60 @@ inline void normalize(T* v, int f) {
...
+	break;
+
+      for (int z = 0; z < f; z++) {
+	iv[z] = iv_sum[z] / ic;
+	jv[z] = jv_sum[z] / jc;
+      }
+
+      if (d_sum < best_d_sum) {
+	std::copy(&iv[0], &iv[f], &best_iv[0]);
+	std::copy(&jv[0], &jv[f], &best_jv[0]);
```
a1k0n: you have a weird 1-space indent here and above

src/annoylib.h
```diff
@@ -437,47 +479,36 @@ template<typename S, typename T, template<typename, typename, typename> class Di
...
+    vector<Node*> children;
+
+    for (size_t i = 0; i < indices.size(); i++) {
+      // TODO: this loop isn't needed for the angular distance, because
+      // we can just split by a random vector and it's fine. For Euclidean
+      // distance we need it to calculate the offset
+      S j = indices[i];
+      Node* n = _get(j);
+      if (n)
+	children.push_back(n);
```
a1k0n: and here...

src/annoylib.h
erikbern: cool will fix

a1k0n:
Oh, there are mixed tabs and spaces, which explains it. Heh.

Anyways, that is awesome. And sort of obvious in retrospect.

src/annoylib.h
a1k0n: Also, you don't set best_d_sum here... so how does this work?

erikbern:
i have a couple of smaller commits coming up, so don't merge yet.

yeah this is quite obvious and i noticed last time i tried it i had a couple of bugs in my code. doing the math i think this is roughly a 2x speedup (i.e. hopefully now it should beat all other ANN libraries)

a1k0n:
I dunno, the small world graph type algos have pretty convincing results...

erikbern:
SW-graph (nmslib) and kgraph are less than 2x faster than Annoy so I hope it's doable :)

src/annoylib.h
```diff
@@ -88,6 +88,60 @@ inline void normalize(T* v, int f) {
...
+    std::copy(&nodes[i]->v[0], &nodes[i]->v[f], &iv[0]);
+    std::copy(&nodes[j]->v[0], &nodes[j]->v[f], &jv[0]);
+    if (cosine) { normalize(&iv[0], f); normalize(&jv[0], f); }
+    T d_sum = 0;
+    
+    for (int iter = 0; iter < 5; iter++) {
+      std::fill(iv_sum.begin(), iv_sum.end(), 0);
+      std::fill(jv_sum.begin(), jv_sum.end(), 0);
+      int ic = 0, jc = 0;
+      for (int l = 0; l < 100 && l < nodes.size(); l++) {
```
a1k0n: ah i see, you sample 100 random nodes here, so there could be a bit of variance here...

erikbern: it's a speedup trick. there's some magic numbers here though, would be nice to avoid

a1k0n:
You still aren't updating best_d_sum on line 134 (apparently github mobile doesn't let me comment on lines directly)

erikbern:
Thanks Andy – that's really crazy, not sure how that ended up outside the PR

src/annoylib.h
```diff
@@ -88,6 +88,60 @@ inline void normalize(T* v, int f) {
...
+      std::fill(iv_sum.begin(), iv_sum.end(), 0);
+      std::fill(jv_sum.begin(), jv_sum.end(), 0);
+      int ic = 0, jc = 0;
+      T d_sum = 0;
+      for (S l = 0; l < 100 && l < (S)nodes.size(); l++) {
+        size_t k = random.index(count);
+        T di = Distance::distance(&iv[0], nodes[k]->v, f),
+          dj = Distance::distance(&jv[0], nodes[k]->v, f);
+        T norm = cosine ? get_norm(nodes[k]->v, f) : 1.0;
+	d_sum += std::min(di, dj);
```
a1k0n: another tab indent here

i recommend :se et, :retab

src/annoylib.h
```diff
@@ -88,6 +88,60 @@ inline void normalize(T* v, int f) {
...
+          for (int z = 0; z < f; z++)
+          jv_sum[z] += nodes[k]->v[z] / norm;
+        }
+      }
+
+      if (ic == 0 || jc == 0)
+        break;
+
+      if (d_sum < best_d_sum) {
+        best_d_sum = d_sum;
```
a1k0n: do you get better results with this in? :)

a1k0n:
LGTM now. this is rad.

src/annoylib.h
```diff
@@ -138,19 +192,11 @@ struct Angular {
...
-    size_t count = nodes.size();
-    size_t i = random.index(count);
-    size_t j = random.index(count-1);
-    j += (j >= i); // ensure that i != j
-    T* iv = nodes[i]->v;
-    T* jv = nodes[j]->v;
-    T i_norm = get_norm(iv, f);
-    T j_norm = get_norm(jv, f);
+    static vector<T> best_iv(f, 0), best_jv(f, 0);
+    two_means<S, T, Random, Angular<S, T, Random> >(nodes, f, random, true, &best_iv[0], &best_jv[0]);
```
a1k0n: do you need to include the parameterized types in <>s here for it to compile? `two_means(nodes, f, ...)` doesn't work?

a1k0n: oh, you do because it can't infer S. why isn't S just `size_t`?

erikbern: not sure, i can try. when i work with templates i really just add/remove "typename" and add/remove template arguments randomly until it compiles :)

src/annoylib.h
```diff
@@ -88,6 +88,60 @@ inline void normalize(T* v, int f) {
...
+    std::copy(&nodes[j]->v[0], &nodes[j]->v[f], &jv[0]);
+    if (cosine) { normalize(&iv[0], f); normalize(&jv[0], f); }
+
+    for (int iter = 0; iter < 5; iter++) {
+      std::fill(iv_sum.begin(), iv_sum.end(), 0);
+      std::fill(jv_sum.begin(), jv_sum.end(), 0);
+      int ic = 0, jc = 0;
+      T d_sum = 0;
+      for (S l = 0; l < 100 && l < (S)nodes.size(); l++) {
+        size_t k = random.index(count);
```
a1k0n: yeah you aren't even using `S` at all here, since k is `size_t`...

src/annoylib.h
erikbern: i think i had it there all the time when i was testing, must have pushed the wrong code for some reason.

src/annoylib.h
a1k0n: ah, it can't infer Distance either, and that's a tougher nut to crack, unless you make two_means a static member of the class instead of having it as a global function.

a1k0n: and that doesn't really help either. so I think you should just get rid of S from two_means, and leave the other arguments in the call.

erikbern: will do. this code would probably benefit from some refactoring but i don't really have the energy/competence

erikbern:
Some benchmarks below.

The problem is I'm getting segfaults when trying to use 400 trees. Fixing this would give even better results since now I'm maxing out at 200 trees. The k-means algorithm uses a bunch more dynamic memory allocation so I will try to rewrite the tree building algorithm to use as little as possible.

Tree building is also 10x slower which is a bit unfortunate but Annoy still is much faster than other algos

image

erikbern:
Turns out since I'm subsampling it's better to do it using SGD rather than in batches. Also seems like the number of attempts doesn't matter at that point – better to increase the number of samples. So code is much simpler now.

a1k0n:
Nice, that makes a lot of sense. Looks good.

a1k0n:
@erikbern Erik, are you going to merge this? It seems like a good idea and it seems to work better..?

erikbern:
i'm planning to merge this but i want to spend a bit more time first – the reason is that I don't fully understand why it works so well, so I'm a bit concerned it will introduce regressions for certain cases.

erikbern:
rebased this, will probably merge this after a bit of cleanup and a few sanity tests to make sure rebasing didn't mess up anything

PR #109, opened 2015-11-07
Close file handlers, simpler reinit code

### LINKED ISSUE #108: "Too Many Files Open" Error - Files Not Being Closed?
I've been using the ANNOY Python API to build and save several index tables in order to load and reuse subsets of these in another script (at most about 400 in a subset at a time). The library is wonderfully fast, but I've been encountering "too many files open" errors after some time. I'm loading each index once, doing stuff with it, then unloading it before the loop continues onwards with the next index table.

Here is a partial code snippet that reproduces the error by reloading the same index table (510 x 64) over and over (run the loop multiple times if the error doesn't show up immediately):

	from annoy import AnnoyIndex
	index_table = AnnoyIndex( 64,metric='angular' )
	for i in xrange(10000):
		index_table.load( index_table_filename )
		print i  # just seeing when it failed
		index_table.unload( index_table_filename )

After some sleuthing into the GitHub source code, I notice that unload() unmaps the file from memory, but doesn't close the file opened using load(). Would it be possible to get file closing functionality incorporated into the Python wrapper somehow please? Or am I missing something vital in using ANNOY?

Thanks for all the hard work on the API!

erikbern:
Resolves #108

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -212,7 +212,6 @@ class AnnoyIndexInterface {
   virtual void add_item(S item, const T* w) = 0;
   virtual void build(int q) = 0;
   virtual bool save(const char* filename) = 0;
-  virtual void reinitialize() = 0;
   virtual void unload() = 0;
   virtual bool load(const char* filename) = 0;
   virtual T get_distance(S i, S j) = 0;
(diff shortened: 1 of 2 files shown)


erikbern:
Seems like tests are failing because GloVe vectors moved, will resolve separately

Issue #100, opened 2015-09-08
double save turns process into zombie on OS X


danielvarga:
A very easy way to turn my python process into a zombie that can't be killed even with -9, and blocks my Terminal tab:

```
from annoy import *
import random

def main():
    f = 10
    t = AnnoyIndex(f)
    for i in xrange(0):
        v = [random.gauss(0, 1) for z in xrange(f)]
        t.add_item(i, v)
    t.build(-1)
    t.save("t.ann")
    t.save("t.ann")
```

In fact, this is all I need:
```
from annoy import *
t = AnnoyIndex(10)
t.save("t.ann")
t.save("t.ann")
```

This happens on an OS X Yosemite. The same thing simply coredumps on Ubuntu.

erikbern:
ok that's not so good. there's some internal state that's not validated. i could add some simple checks

danielvarga:
Please do! Thanks a lot! Amusingly, this weird zombie process somehow interferes with the working of live processes, for example makes Preview dead slow, reproducibly. (Takes two minutes to open a document.) Only a full system restart helps. I'm wondering how can that even happen.

PR #110, opened 2015-11-07
Fix double save bug in #100

erikbern:
Fix double save bug in #100


a1k0n:
:rocket:

erikbern:
will fix the glove stuff separately

Issue #112, opened 2015-11-11
About distances


anhldbk:
Hi there. I'm trying to use Annoy to get distances information using 'angular' (or cosine distance). But I can't get it right. 

My code is below: 
```python
from annoy import AnnoyIndex
import random

f = 40
t = AnnoyIndex(f)  # Length of item vector that will be indexed
for i in xrange(1000):
    v = [random.gauss(0, 1) for z in xrange(f)]
    t.add_item(i, v)

t.build(10)  # 10 trees

v = [random.gauss(0, 1) for z in xrange(f)]
data = t.get_nns_by_vector(v, 5, include_distances=True)  # will find the 5 nearest neighbors
print data
```
I expected to have an output with distances lesser than 1. But what I got is:

```text
([227, 987, 542, 302, 407], [0.9761863350868225, 1.085146427154541, 1.0856424570083618, 1.1076160669326782, 1.116296648979187])```

Would you pls tell me what is wrong?

erikbern:
Cosine distance can be between 0 and 2. The distance between x and -x should be exactly 2

http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.spatial.distance.cosine.html

I will add a unit tests though

anhldbk:
@erikbern Thank you erik

harsh157:
Shouldn't the distance range be 0 to 4? as it is computing 2*(1-cos). That is if we have vectors in opposing directions also.

MrChrisJohnson:
Why 2*(1-cos)? Distance should just be 1-cos which is equivalent to 1 - u*v / ||u|| * ||v||. So, distance would range from 0 (angle of 0) to 2 (angle of 180). See the link that @erikbern links to earlier.

Irtza:
@erikbern @MrChrisJohnson Can we convert to cosine similarity using cos_sim = 1-cos_distance ?

erikbern:
cosine_distance = sqrt(2-2*cos_similarity)

so cosine_similarity = 1 - cosine_distance^2/2

PR #113, opened 2015-11-11
Tests for angular distances and fix small bug

erikbern:
Added some tests to verify #112 and realized there was a bug causing distances for identical vectors to be returned as `nan` sometimes


erikbern:
This is weird, if I do sqrt(fmax(x, 0)) where x = -0.0 then it seems like it returns a small positive value. Not sure what's going on

PR #116, opened 2015-11-18
Changed templatization of Angular and Euclidean to be on the function…

dcdillon:
…s not the class.

This is already in rcppannoy (for the last day so it hasn't had too much time to burn in).  It simply makes the interface to `Angular` and `Euclidean` a bit easier for the end user.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -89,8 +89,8 @@ inline void normalize(T* v, int f) {
 }
 
 
-template<typename S, typename T, class Random>
 struct Angular {
+  template<typename S, typename T>
   struct ANNOY_NODE_ATTRIBUTE Node {
     /*
      * We store a binary tree where each node has two things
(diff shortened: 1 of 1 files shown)


erikbern:
thanks!

src/annoylib.h
```diff
@@ -233,10 +243,11 @@ template<typename S, typename T, template<typename, typename, typename> class Di
    * We create a tree like this q times. The default q is determined automatically
    * in such a way that we at most use 2x as much memory as the vectors take.
    */
-protected:
-  typedef Distance<S, T, Random> D;
-  typedef typename D::Node Node;
+public:
+  typedef Distance D;
+  typedef typename D::template Node<S, T> Node;
```
a1k0n: Wow, I've never seen `template` used in this construct. What does it mean here?

dcdillon: It's somewhat like when you have to use typename on nested types.  It just tells the compiler that D::Node is a template (and not a variable, standard type, etc.)

PR #127, opened 2016-01-04
Python Docstrings

Joshua-Chin:
This adds docstrings to the Python bindings. This also fixes an error in the documentation of the `unload` method. The docs say that it takes a single argument, but the code doesn't accept any.


erikbern:
Great! I can probably write a macro that pulls out the docstrings and writes it into the README (or maybe just use readthedocs)

PR #137, opened 2016-02-17
fix for duplicate roots

erikbern:
noticed a minor bug causing roots to be searched twice (useless)

currently seeing some weird segfaults on my computer triggered by something else, investigating in parallel

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -403,6 +403,9 @@ public:
         break;
       }
     }
+    // hacky fix: since the last root precedes the copy of all roots, delete it
+    if (_roots.size() > 1 && _get(_roots.front())->children[0] == _get(_roots.back())->children[0])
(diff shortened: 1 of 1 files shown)


a1k0n:
Whoa. Is one of the roots being added twice, or is the scan just seeing it twice somehow?

erikbern:
the latter

erikbern:
it would blow up the search time by a factor of (trees + 1) / trees so not that much

a1k0n:
Hm. I don't understand how... but that code isn't the clearest...

erikbern:
i copy the roots to the end of the index. but if the last item of the index was already a root then you end up having n+1 roots at the end of the index. the way annoy finds the roots is it just takes the last few nodes that has the biggest number of descendants. so that would include the last root twice (and all the other roots once)

PR #138, opened 2016-02-20
put stuff on heap to avoid segfaults

erikbern:
Had some issues where unit tests would throw segfaults after success


a1k0n:
now it's on the stack, not the heap. it was in bss. Either way if it was segfaulting then something isn't right...

Have you tried valgrind?

erikbern:
you are right that my description was wrong

i think the problem was that static objects are only initialized once. so if f would increase, then the second time it wouldn't be resized properly

i'll probably remove all dynamic memory allocations in a separate PR

erikbern:
i tried valgrind but it went nuts because annoy does all this magic stuff with pointer offsets. so wasn't super helpful.

PR #154, opened 2016-05-01
ensure item indices are inside range, add test

starius:
The following commands used to cause segmentation fault
when called with negative or out of range index:

  * add_item (only applies to negative indices)
  * get_distance
  * get_nns_by_item
  * get_item_vector


erikbern:
cool

PR #164, opened 2016-07-16
Support for threads

erikbern:
This is partly so we can max out the CPU cores while running ann-benchmarks (since other libraries might be multithreaded)

See https://github.com/erikbern/ann-benchmarks/pull/28#issuecomment-231733072


erikbern:
runtime of the new test went from 15s to 6s after the fix

the speedup should be a lot more though since there's some serial overhead (Amdahl's law)

PR #166, opened 2016-07-25
cpp precision example and compilation, no unit test here

csanhuezalobos:
Here, I include the precision example written in cpp. Also, a tiny script for compilation purposes.


erikbern:
nice! will see if there's an easy way to run this as a part of unit tests

csanhuezalobos:
Really small contribution, but it was part of my learning process.

PR #171, opened 2016-08-26
Fix distance inconsistency for angular distance

erikbern:
Now normalizing distance for `get_distance` meaning the return values will be changing – I hope not a lot of people use `get_distance`! But I think some breakage is worth in order to get consistency

Also fixed a typo in tests: Equals -> Equal

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,7 +34,7 @@ Annoy was built by `Erik Bernhardsson <http://www.erikbern.com>`__ in a couple o
 Summary of features
 -------------------
 
-* Euclidean distance or cosine distance (where cosine distance uses Euclidean distance of normalized vectors which equals 2-2*cosine similarity)
+* Euclidean distance or cosine distance, where cosine distance uses Euclidean distance of normalized vectors = sqrt(2-2*cos(u, v))
 * Works better if you don't have too many dimensions (like <100) but seems to perform surprisingly well even up to 1,000 dimensions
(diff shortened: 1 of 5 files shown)


erikbern:
interestingly this breaks the Lua tests – will update them

README.rst
```diff
@@ -34,7 +34,7 @@ Annoy was built by `Erik Bernhardsson <http://www.erikbern.com>`__ in a couple o
 Summary of features
 -------------------
 
-* Euclidean distance or cosine distance (where cosine distance uses Euclidean distance of normalized vectors which equals 2-2*cosine similarity)
+* Euclidean distance or cosine distance, where cosine distance uses Euclidean distance of normalized vectors = sqrt(2-2*cos(u, v))
```
andy256: Hah, I never noticed that before but now that I think about it, this makes sense.

This doesn't actually affect any accuracy benchmarks does it?

erikbern: no – doesn't change anything except the values returned by `get_distance`

still on the fence about merging this since it might break existing users, but I think i'll merge it into master and wait a few weeks to publish it to pypi

a1k0n: sounds good to me.

erikbern: haha what's up with two github accounts?

a1k0n: argh, was logged into the wrong one on my phone, haha. ignore that!

PR #178, opened 2016-10-10
Added required methods and interfaces for unbuild() method

### LINKED ISSUE #174: What's the recommended way of adding new items after build() has been called?
Right now I have to create a new instance, retrieve all the feature vectors by iterating over all the items in the old instance add them to the new one and finally unload the old one. Is there any better way?

It would be really nice to have something like an "unbuild()" method built in which would do that without having to temporarily double the memory requirements.

Quasimondo:
This resolves #174 - at least for C++ and the Python wrapper

### DIFF
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -250,6 +250,19 @@ py_an_build(py_annoy *self, PyObject *args) {
 
 
 static PyObject *
+py_an_unbuild(py_annoy *self, PyObject *args) {
+  if (!self->ptr) 
+    Py_RETURN_NONE;
+  
+  Py_BEGIN_ALLOW_THREADS;
+  self->ptr->unbuild();
+  Py_END_ALLOW_THREADS;
+
+  Py_RETURN_TRUE;
+}
+
+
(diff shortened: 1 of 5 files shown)


erikbern:
nice!

can you add a unit test for this? ideally something that builds and unbuilds a bunch of times

src/annoymodule.cc
```diff
@@ -250,6 +250,19 @@ py_an_build(py_annoy *self, PyObject *args) {
 
 
 static PyObject *
+py_an_unbuild(py_annoy *self, PyObject *args) {
+  if (!self->ptr) 
+    Py_RETURN_NONE;
+  
+  Py_BEGIN_ALLOW_THREADS;
```
erikbern: cool yeah i don't think you need to go crazy with it, just run `build` and `unbuild` a bunch of times and make sure things don't crash :)

erikbern:
waiting for a few unit tests for this and i'm happy to merge

erikbern:
looks good for now! thanks!

PR #186, opened 2017-02-25
Avoid referencing out of bounds elements

lisitsyn:
Current way of obtaining bounds for `std::partial_sort` leads to out of bounds assertion error in `std::vector`. This patch replaces `access by index + dereferencing` with getting pointer for 0th element and then advancing it forward. The latter approach is equivalent but safe.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -596,7 +596,7 @@ protected:
 
     size_t m = nns_dist.size();
     size_t p = n < m ? n : m; // Return this many items
-    std::partial_sort(&nns_dist[0], &nns_dist[p], &nns_dist[m]);
+    std::partial_sort(nns_dist.begin(), nns_dist.begin() + p, nns_dist.end());
     for (size_t i = 0; i < p; i++) {
       if (distances)
         distances->push_back(D::normalized_distance(nns_dist[i].first));


erikbern:
I don't understand how this can lead to an out of bounds assertion since p is always less than or equal to the length of the array?

lisitsyn:
@erikbern I was hunting down exactly what @a1k0n said.

Usual `begin` and `end` would be better, indeed. Let me know if you prefer these.

erikbern:
I still don't understand what bug this solves. Can you send a reproducible case?

a1k0n:
If you have an STL that does bounds checking (Microsoft compiler maybe?)
then vector::operator[] will throw an exception on that line.

erikbern:
Hm weird... `&v[p]` doesn't actually access `v[p]` though, it's just pointer arithmetic. It shouldn't invoke `operator[]` afaik.

andy256:
It's not just pointer arithmetic, because nns_dist is a vector and it overloads []

andy256:
That's why you can fix it by referencing [0] and THEN doing pointer arithmetic

erikbern:
cool i guess that makes sense

PR #187, opened 2017-03-02
enable annoy seeding

yonromai:
@erikbern

### DIFF
--- a/annoy/__init__.py
+++ b/annoy/__init__.py
@@ -130,3 +130,9 @@ class AnnoyIndex(Annoy):
         Returns the number of items in the index.
         """
         return super(AnnoyIndex, self).get_n_items()
+
+    def set_seed(self, seed):
+        """
+        Sets the seed of Annoy's random number generator.
+        """
+        return super(AnnoyIndex, self).set_seed(seed)
--- a/annoy/__init__.py
+++ b/annoy/__init__.py
@@ -130,9 +130,3 @@ class AnnoyIndex(Annoy):
         Returns the number of items in the index.
         """
         return super(AnnoyIndex, self).get_n_items()
-
-    def set_seed(self, seed):
-        """
-        Sets the seed of Annoy's random number generator.
-        """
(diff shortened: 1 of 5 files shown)


andy256:
You'll probably want to extend the RandRandom strict with this, and maybe implement it for the other (boost?) RANG that's available...

andy256:
I hate autocorrect. Struct, RNG

yonromai:
@andy256: perhaps a simple seeding would do?

yonromai:
@andy256: Is this what you mean? It seems like the other RNGs (KissXYRandom) are already seeded

a1k0n:
I was suggesting keeping the API to set the seed, and just pass it to the RNG... but I see now that it's part of the constructor, so that might be tricky. @erikbern thoughts?

erikbern:
what's the point of this?

andy256:
Reproducible builds?

yonromai:
@andy256: is this more like what you had in mind?
(don't worry I'll squash if we want to merge...)

erikbern:
i guess, but then what are reproducible builds useful for? :)

yonromai:
lol, CI/testing are totally overrated ;)
@erikbern do you think this change would impact perf of annoy?

erikbern:
i mean that if you have flakiness in your tests then you should fix that... don't rely on exact comparisons between the results arrays for instance, allow some percentage of missing values

see the unit tests for annoy – https://github.com/spotify/annoy/blob/master/test/annoy_test.py#L138

I don't have a problem with this PR just pointing out I think you're solving it in the wrong place

yonromai:
actually AFAIC it's not for testing, but allowing to seed the RNG would be nice to make our pipelines idempotent. 
But I don't have superstrong feelings about this, so feel free to close the PR if you disagree.

a1k0n:
Yeah, it would be a nice property if two speculatively executed tasks ended up with the same result.

erikbern:
no i think seeding is generally good to have... for some things (like train/test split) it's crucial.

Issue #188, opened 2017-03-03
Not having consistent results after setting seed


gg2572:
Hi Team,
I hope you're all doing well. Thank you for all hard works you put to develop Annoy.
I've noticed that the most recent commit adds the functionality to set seed, and this is really good for making results reproducible. However, when I run the program with the follow code, it does not produce the same results. Would anyone of you can help me to look at it? 

Thank you very much. I appreciate it.

Best,

Gan

```
import numpy as np
from annoy import AnnoyIndex

X = np.random.rand(100000, 60)
Y = np.random.rand(500, 60)

annoy1 = AnnoyIndex(60)
annoy1.set_seed(100)
for i in range(X.shape[0]):
    annoy1.add_item(i, X[i, :])

annoy1.build(10)

annoy2 = AnnoyIndex(60)
annoy2.set_seed(100)
for j in range(X.shape[0]):
    annoy2.add_item(j, X[j, :])

annoy2.build(10)

result1 = []
result2 = []

for k in range(Y.shape[0]):
    print "annoy1", annoy1.get_nns_by_vector(Y[k, :], 3)
    result1 += annoy1.get_nns_by_vector(Y[k, :], 3)
    print "annoy2", annoy2.get_nns_by_vector(Y[k, :], 3)
    result2 += annoy2.get_nns_by_vector(Y[k, :], 3)
```

ijklr:
Hi Gan, I ran your python, it produced the same results for me. Where do
you see the difference?

gg2572:
Hi Yi Cheng,
Thank you for your reply!
Yes, here's what I get.
```annoy1 [23559, 25499, 49827]
annoy2 [23559, 25499, 49827]
annoy1 [97178, 58504, 30602]
annoy2 [97178, 58504, 30602]
annoy1 [16834, 15731, 15958]
annoy2 [16834, 15731, 15958]
annoy1 [13066, 48258, 14626]
annoy2 [40847, 9026, 28475]
annoy1 [82735, 37999, 56475]
annoy2 [49271, 50974, 45406]
annoy1 [8258, 99461, 3842]
annoy2 [74761, 17914, 42611]```

Sometimes I get the same but sometimes they're different.

gg2572:
So I tried with the same code with a different `numpy` version (1.11.3) and now they're the same. Trying to figure out why the latest `numpy` (1.12.0) is causing the problem.

yonromai:
[...]

PR #189, opened 2017-03-09
add test for seeding

erikbern:
Trying to reproduce #188


test/annoy_test.py
```diff
@@ -523,3 +523,25 @@ def test_threads(self):
         def query_f(j):
             i.get_nns_by_item(1, 1000)
         pool.map(query_f, range(n))
+
+
+class SeedTest(TestCase):
+    def test_seeding(self):
+        print('seeding')
```
erikbern: oops should have removed this line

whatever

Issue #194, opened 2017-03-29
index with single vector after save


tpeng:
Hi,

I noticed following code return wrong result:

```python
from annoy import AnnoyIndex

a = AnnoyIndex(3)
a.add_item(0, [1, 0, 0])
a.build(10)
a.save('1.ann')
print(a.get_nns_by_vector([1, 0, 0], 3, include_distances=True)) # ([1, 2, 3], [0.0, 0.0, 0.0])
```

but it works well  without `a.save('1.ann')` (return `([0], [0.0])`)

any thoughts?

erikbern:
this sounds like a bug. let me add a test case for it and see if i can reproduce. thanks!

erikbern:
I'm not able to reproduce this – see #196

did you use the latest version?

tpeng:
hmm, apologies, i use 1.8.0.
checking the commit history, looks like https://github.com/spotify/annoy/commit/c33b84909aa96b75a055fe78b909ef53a9f1ba5c fixed it. closing it.

erikbern:
ok awesome :)

PR #196, opened 2017-03-30
add unit test for #194

erikbern:


erikbern:
not sure why this one was sitting here so long :)

PR #204, opened 2017-06-14
Fixed examples/precision_test.cpp

natanlaverde:
Fix example to use KISS random number generators instead of RandRandom


erikbern:
cool

Issue #190, opened 2017-03-12
New set_seed not always producing consistent results


pommedeterresautee:
I was trying to have reproductible results with Annoy and discovered the `set_seed` feature has been added merged very recently (https://github.com/spotify/annoy/commit/2de9fd3eee45ef83b1e8d6ed372edc370e503a5c).

In my tests, when I compare several times the retrieval for 2 models based on a large dataset (like 1 million random vectors) of 300d for instance, sometimes I get the same results, sometimes not.
Seems to happen more when the dataset is large that when it is small and dimensions > 100.
FYI results are always perfectly exact with 1000 vectors and 50d (tried many times).

When the test is failing, results retrieved from both models are "very similar" but with the same seed I expect exactly the same resutls. Seems to be an issue with `set_seed` only.

Any idea @yonromai @erikbern ?

FYI my code is in Scala, just calling the C++ with JNI, nothing done in between.

```scala
  val seed = 42
  Random.setSeed(seed)
  val numberOfVectors = 1000000
  val dimensions = 300
  val trees = 30
  val vectors = Array.fill(numberOfVectors)(Array.fill(dimensions)(Random.nextGaussian().toFloat)).zipWithIndex

  val annoy1 = AnnoyInstance(new File(""), dimensions, seed = seed, loadModel = false, verbose = true)
  for ((vector, index) ← vectors) {
    annoy1.addItem(index, vector)
  }
  annoy1.build(trees)
  val annoy2 = AnnoyInstance(new File(""), dimensions, seed = seed, loadModel = false, verbose = true)
  for ((vector, index) ← vectors) {
    annoy2.addItem(index, vector)
  }
  annoy2.build(trees)
  assert(annoy1.getNItems == numberOfVectors)
  assert(annoy2.getNItems == numberOfVectors)
  for (_ ← 1 to 500) {
    val elementToTest = Random.nextInt(numberOfVectors)
    assert(annoy1.getNnsByItem(elementToTest, 100).sameElements(annoy2.getNnsByItem(elementToTest, 100)))
  }
```

erikbern:
I'm not able to reproduce this. See https://github.com/spotify/annoy/blob/master/test/annoy_test.py#L528

I increased this to 100k vectors, 300D, 30 trees

Ran it three times, passes every time

pommedeterresautee:
Python version works also for me (tried with 10 millions 10 times).
I close the issue for now.
Will reopen if I find something related to the C++ part.

[...]

PR #205, opened 2017-06-15
Add compilation flag to preserve deterministic behaviour

rshest:
This commit fixes possibly non-deterministic building of the index when running on an Intel processor (see issues #190 and #188) by adding compilation flag `-fno-associative-math`.

This could happen even when the seed is set, running the same binary on exactly the same inputs. The culprit is the compiler setting `-ffast-math`, which in turn enables `-funsafe-math-optimizations`, which in turn enables `-fassociative-math`, which allows compiler to treat math operations as associative, so that order of summation in the function `margin` (that computes dot product) can be arbitrary.

It apparently can also be arbitrary between different runs of the same program on the same inputs, [in particular when loops are vectorized:](
http://techdoc.dartmouth.edu/discovery/wp-content/uploads/FP_Consistency_12.pdf)

> Slightly different results were observed when re-running the same (non-threaded)
> binary on the same data on the same processor.
> This was caused by variations in the starting address and alignment of the global
> stack, resulting from events external to the program. The resulting change in local
> stack alignment led to changes in which loop iterations were assigned to the loop
> prologue or epilogue, and which to the vectorized loop kernel. This in turn led to
> changes in the order of operations for vectorized reductions (i.e., reassociation).

("slightly different" in case of the `margin` function  can be enough to build the tree in a different way; the result is used right after that in the `side` function, which takes a binary decision)

Enabling back the computation order guarantee does not seem to visibly affect performance, but it brings back reproducibility, which looks like a fair tradeoff.

Note that in my case (Intel i7-4790K, Linux Mint 18, annoy v.1.8.3) the existing unit test `SeedTest` already intermittently failed (once in 10-ish runs). Bumping the number of vectors to 500000 and number of trees to 50 made test failure almost certain.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,8 @@ setup(name='annoy',
         Extension(
             'annoy.annoylib', ['src/annoymodule.cc'],
             depends=['src/annoylib.h', 'src/kissrandom.h', 'src/mman.h'],
-            extra_compile_args=['-O3', '-march=native', '-ffast-math'] + travis_extra_compile_args,
+            extra_compile_args=['-O3', '-march=native',
+                                '-ffast-math', '-fno-associative-math'] + travis_extra_compile_args,
         )
       ],
       long_description=long_description,


erikbern:
@rshest hey sorry forgot about your PR. I just rebased it on latest master. does it look OK?

rshest:
@erikbern Looks good, thanks a lot!

PR #206, opened 2017-06-27
Manhattan

natanlaverde:
Added support to Manhattan metric and Python tests to Manhattan metric.


erikbern:
very cool! i think this makes a lot of sense. merging it for now since i'm going to forget to do the more thorough review in any reasonable time, but i'm hoping to get back to it soon with some comments!

PR #213, opened 2017-07-24
remove a bunch of python glue

erikbern:
...and move it in to the C++ layer


erikbern:
will push to pypi shortly

PR #224, opened 2017-09-30
don't return nodes that were never added (holes)

### LINKED ISSUE #223: Annoy less resilient to id holes than before
Hey @erikbern, 

TL;DR: `1.9.4` is a lot less robust to holes in the id space than `1.3.2`.

--

After upgrading annoy to `1.9.4` for DW, we got a pretty significant drop in metrics. 

After investigations, it seems that compared to older versions of annoy, the latest version is not very robust to holes in the IDs (which we have for DW because of some multistage filtering logic). 

In an experiment where I compared building a tree without holes (id: 0 to n-1) with 30% of holes in the ids, I found that annoy `1.3.2` got a 3% drop in perf (cos sim of nn) where annoy `1.9.4` perf dropped by 36%.

Here are the numbers for mean cos sim of a bunch of nn queries:
```
	old_annoy	new_annoy	old_holes_annoy	new_holes_annoy
count	5000.000000	5000.000000	3499.000000	3520.000000
mean	0.135056	0.132766	0.140376	0.180412
std	0.013602	0.013103	0.017448	0.017294
min	0.099506	0.096045	0.091715	0.130184
25%	0.125277	0.123374	0.128233	0.168452
50%	0.134288	0.131794	0.138153	0.179342
75%	0.143350	0.141121	0.150459	0.191101
max	0.197450	0.189082	0.218467	0.257536
```

(If you want the code to reproduce this I can open source it)

Were you aware of this issue?

erikbern:
Fixes #223

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -363,8 +363,10 @@ public:
       if (_verbose) showUpdate("pass %zd...\n", _roots.size());
 
       vector<S> indices;
-      for (S i = 0; i < _n_items; i++)
-        indices.push_back(i);
+      for (S i = 0; i < _n_items; i++) {
+	if (_get(i)->n_descendants >= 1) // Issue #223
+	  indices.push_back(i);
+      }
 
       _roots.push_back(_make_tree(indices));
     }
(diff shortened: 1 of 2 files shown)


ksindi:
:shipit: on tests

erikbern:
something is broken on travis with python 3.5, going to ignore for now

PR #225, opened 2017-09-30
run python 3.6 instead of 3.5

erikbern:


erikbern:
still not working... not sure what's up. i hate tox, might remove it

PR #231, opened 2017-10-25
Adding support for binary vectors under hamming distance

maumueller:
## Purpose
This PR adds support for binary vectors (stored in 64-bit chunks) under Hamming distance. Instead of random projections, it splits by projecting a vector to a single bit at a random position, which is known to be optimal in the locality-sensitive hashing world. Distance computations are done xor'ing and counting ones in the result using built-in `popcount` intrinsics.

These changes made some refactoring necessary in the query code w.r.t. the PQ. The initial value and the priority that is associated with an element is now configurable from within the distance measure struct. The compiler was able to inline these calls on my setup, yielding identical code/performance. 

## Performance

Below some screenshots comparing the performance. 
screen shot 2017-10-25 at 14 47 02

This is from a binary version of the SIFT dataset. Both euclidean and hamming distance versions allow to get recall close to 1, hamming distance is a bit faster. Original version from spotify/annoy as fast as the updated version with the changed PQ interface.

screen shot 2017-10-25 at 14 47 12

This is from a binary version of the NYtimes dataset. Projecting to a single bit works much better quality wise.

## Bottom line

I don't know much about the wrappers to annoy. The representation of data points from the python side would be what `numpy.packbits` with `dtype=uint64` produces.

Looking forward to hearing your thoughts on this PR.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -699,7 +699,7 @@ protected:
       search_k = n * _roots.size(); // slightly arbitrary default value
 
     for (size_t i = 0; i < _roots.size(); i++) {
-      q.push(make_pair(Distance::pq_initial_value(), _roots[i]));
+      q.push(make_pair(Distance::template pq_initial_value<T>(), _roots[i]));
     }
 
     vector<S> nns;
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -178,8 +178,107 @@ struct Angular {
     // so we have to make sure it's a positive number.
     return sqrt(std::max(distance, T(0)));
   }
+  template<typename T>
+  static inline T pq_distance(T distance, T margin, int child_nr) {
+    if (child_nr == 0)
+      margin = -margin;
+    return std::min(distance, margin);
+  }
+  template<typename T>
+  static inline T pq_initial_value() {
+    return numeric_limits<T>::infinity();
+  }
+  static inline int chunkcount(int f) {
+    return f;
+  }
   static const char* name() {
     return "angular";
+  }
+};
+
+struct Hamming {
+
+  template<typename S, typename T = int64_t>
+  struct ANNOY_NODE_ATTRIBUTE Node {
+    S n_descendants;
+    S children[2];
+    T v[1];
+  };
+
+  static inline size_t chunksize() {
+    return 64;
+  }
+  static const size_t max_iterations = 20;
+
+  template<typename T = int64_t>
+  static inline T pq_distance(T distance, T margin, int child_nr) {
+    return distance - (margin != child_nr);
+  }
+
+  template<typename T = int64_t>
+  static inline T pq_initial_value() {
+    return 0;
+  }
+  template<typename T = int64_t>
+  static inline T distance(const T* x, const T* y, int f) {
+    size_t dist = 0;
+    for (size_t i = 0; i < chunkcount(f); i++)
+    {
+      dist += __builtin_popcountll(x[i] ^ y[i]);
+    }
+    return dist;
+  }
+  template<typename S, typename T = int64_t>
+  static inline bool margin(const Node<S, T>* n, const T* y, int f) {
+    T chunk = n->v[0] / chunksize();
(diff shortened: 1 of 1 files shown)


erikbern:
This is excellent – really appreciate the addition. I will review in the next few days. Should add a few unit tests in Python as well

erikbern:
Random thought but I guess it would be very easy to support Jaccard similarity as well based on the binary vector? Just an idea

erikbern:
Hoping to take a look at this during the weekend

maumueller:
Conceptually that wouldn't be a problem. However, we need to use something like MinHash to do the split in a node.

erikbern:
true that's a good point. how does the split algo work for hamming distance?

maumueller:
It looks at a random position in the bit string, elements with a 0 go to left, elements with a 1 go to the right.

erikbern:
only one single random position? think you can probably improve on that algorithm

maumueller:
Well, it worked much better than a projection (say, take the inner product with a random {-1,1} vector and split by <0 and >= 0). It’s also what the standard LSH hash family hamming distance would do. You could look at multiple positions at once, but then I didn’t know what a good split criterion is.

erikbern:
ok, interesting. i would have assumed random projection would be better.

src/annoylib.h
```diff
@@ -178,11 +178,111 @@ struct Angular {
...
+  static inline bool margin(const Node<S, T>* n, const T* y, int f) {
+    T chunk = n->v[0] / chunksize();
+    return (y[chunk] & (static_cast<T>(1) << (chunksize() - 1 - (n->v[0] % chunksize())))) != 0;
+  }
+  template<typename S, typename T = int64_t, typename Random>
+  static inline bool side(const Node<S, T>* n, const T* y, int f, Random& random) {
+    return margin(n, y, f);
+  }
+  static inline int chunkcount(int f) {
+    return f / chunksize() + 1;
```
erikbern: this should be `f / chunksize()` rounded up right? so i think it should be
`(f + chunksize() - 1) / chunksize()` instead?

erikbern:
Is there a reason you chose to preserve the meaning of `f` as the dimensionality rather than the number of vector "slots" – I think in the latter case if you let `f` basically represent the "chunk count" then you don't have to introduce the extra `chunksize` and `chunkcount` helpers. I think that would be marginally simpler to deal with in the code.

My biggest question design wise is whether the Python layer should accept 64 bit integers as vectors, or binary vectors. The former is far more efficient and leads to simpler code, but the latter might be a bit more logical from a user point of view. I might start by just doing the 64 bit thing since that's simpler

erikbern:
actually after looking into it a bit more, i think the problem is how to deal with the python layer – atm the python object contains an `AnnoyIndexInterface<int32_t, float>* ptr` pointer, so dealing with any other types is problematic. not sure how to solve it

maumueller:
>Is there a reason you chose to preserve the meaning of f as the dimensionality rather than the number of vector "slots" – I think in the latter case if you let f basically represent the "chunk count" then you don't have to introduce the extra chunksize and chunkcount helpers. I think that would be marginally simpler to deal with in the code.

No reason except from me being unable to predict whether `f` in its current meaning might be useful at another point. I don't like the `chunksize` and `chunkcount` helpers either, but wanted to change as few of the original code logic as possible.

> one idea would be to create a wrapper object that implements AnnoyIndexInterface<int32_t, float> but contains a pointer to a AnnoyIndex<int32_t, int64_t, Hamming, Kiss64Random> and does the translation to/from binary vectors. probably the easiest approach.

I have no experience how to write these Python wrappers, but I guess you want to have a switch `metric = "hamming"` to `AnnoyIndex`, and not a different type of index, say `AnnoyIndexBitvector`. The latter would allow for a cleaner separation, but the former is easier to use from a user perspective. So, hiding it inside the wrapper may be ok. (Although it feels a bit strange conceptually, because the float variant and the bit variant are on the same level.)

src/annoylib.h
maumueller: Yes, you are right! Fixing this.

PR #235, opened 2017-10-31
Adding Python 2.7 support for Windows

tjrileywisc:
Added some compiler checks for MSVC 2008 and 'reverted' #3 in the case of the MSVC 2008 compiler (and for some reason when debugging with any MSVC compiler, std::copy isn't working).

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -27,7 +27,12 @@
 #include <sys/types.h>
 #include <fcntl.h>
 #include <stddef.h>
+#if defined(_MSC_VER) && _MSC_VER == 1500
+typedef unsigned char     uint8_t;
+typedef signed __int32    int32_t;
+#else
 #include <stdint.h>
+#endif
 
 #ifdef _MSC_VER
 #define NOMINMAX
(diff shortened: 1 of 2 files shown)


src/annoylib.h
```diff
@@ -530,9 +535,17 @@ template<typename S, typename T, typename Distance, typename Random>
       Node* m = _get(item);
       m->n_descendants = (S)indices.size();
 
-      // Using std::copy instead of a loop seems to resolve issues #3 and #13,
-      // probably because gcc 4.8 goes overboard with optimizations.
+    // Using std::copy instead of a loop seems to resolve issues #3 and #13,
+    // probably because gcc 4.8 goes overboard with optimizations.
+
+    // (but copy not cooperating with MSVC in certain conditions)
+#if (defined(_MSC_VER) && defined(_DEBUG)) || (defined(_MSC_VER) && (_MSC_VER == 1500))
```
erikbern: IIRC there's actually two places that uses `std::copy` in the code.

might be worth making this a macro btw so that we don't clutter the code with too much compiler specific stuff

erikbern:
very nice – see comment i left!

erikbern:
happy to merge this if you address my comment about `std::copy`

tjrileywisc:
I've been trying to figure out a solution- my C++ knowledge is pretty weak so it will probably take me a while.

I'm actually trying to figure out why the crash is happening in the first place instead of using the loop. I think it might be because it's indexing out of bounds. 

Maybe some reallocation needs to be done? m->children has length of 2 but at the time of the crash there are more than 2.

erikbern:
we could probably actually replace it with `memcpy` and it should work on both platforms... let me see

erikbern:
Take a look at #236 – it might work and remove the need for MSVC-specific workarounds

erikbern:
closing this in favor of #238

PR #238, opened 2017-11-04
Windows Python 2.7 support

erikbern:
This should supersede #235

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -27,7 +27,12 @@
 #include <sys/types.h>
 #include <fcntl.h>
 #include <stddef.h>
+#if defined(_MSC_VER) && _MSC_VER == 1500
+typedef unsigned char     uint8_t;
+typedef signed __int32    int32_t;
+#else
 #include <stdint.h>
+#endif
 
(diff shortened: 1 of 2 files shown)


erikbern:
merging this – @tjrileywisc can you see if this works?

tjrileywisc:
@erikbern I'm afraid that didn't work. I'm seeing python crashes in 3.6 and 2.7 now, trying to figure out why.

I'm able to compile the precision_test.cpp example though in MSVC and it runs fine, so I think the problem is in the python module specifically.

erikbern:
hm that's annoying. sorry i really thought my fix would resolve the issue

do you see any compiler warnings? anything else that could indicate what the issue is?

erikbern:
Finding some evidence that MSVC adds memory checks to `memcpy`: http://www.informit.com/articles/article.aspx?p=2036582&seqNum=6

maybe add `#pragma runtime_checks("s", off)` at the top of the file?

farizrahman4u:
@erikbern 

> hm that's annoying.

I see what you did there.

PR #239, opened 2017-11-04
disable runtime memcpy check for visual studio

erikbern:
see discussion in #238

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -49,6 +49,9 @@ typedef signed __int32    int32_t;
 #include <queue>
 #include <limits>
 
+// Needed for Visual Studio to disable runtime checks for mempcy
+#pragma runtime_checks("s", off)
+
 // This allows others to supply their own logger / error printer without
 // requiring Annoy to import their headers. See RcppAnnoy for a use case.
 #ifndef __ERROR_PRINTER_OVERRIDE__
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -448,8 +448,10 @@ public:
 
   bool load(const char* filename) {
     _fd = open(filename, O_RDONLY, (int)0400);
-    if (_fd == -1)
+    if (_fd == -1) {
+      _fd = 0;
       return false;
+    }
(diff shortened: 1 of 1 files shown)


erikbern:
@tjrileywisc let me know if this works!

tjrileywisc:
Still didn't work I'm afraid. Looks like it's crashing python in the same place. It's crashing when running the nosetests. This is the last output just before the crash:

finding nbs for 161
finding nbs for 22055
finding nbs for 49154
finding nbs for 54985
finding nbs for 36080
finding nbs for 25961
finding nbs for 47264
limit: 10        precision:  13.56% avg time: 0.000068s
limit: 100       precision:  21.06% avg time: 0.000095s
limit: 1000      precision:  56.51% avg time: 0.000311s
limit: 10000     precision:  97.12% avg time: 0.002118s
.........

tjrileywisc:
Okay, might have a signal here. Finally got to a point where I could use MSVC debug symbols to find out where exactly it's crashing:

`AnnoyIndex<int,float,Angular,Kiss64Random>::unload() Line 440`

... which I think sounds about right, from the output above. I think all of the tests run and then it crashes when finishing up.

It's crashing at this function:
`munmap(_nodes, size);`

`size` appears to be zero, because `_n_nodes` is zero.

erikbern:
`munmap` is in `unload` which is run by the destructor. But it's only run if `_fd` is nonzero.

The only reason I can see that `_fd` would be nonzero would be if we try to load a file then fail. In that case it ends up being `-1`. Pushing a fix as a long shot – let's see if it helps

erikbern:
It would be useful if you could run nosetests on each test file separately to see which one triggers the failure. Eg `python nosetests test/index_test.py` (I actually suspect that's the test that breaks)

erikbern:
merging this for now... should be harmless at worst

erikbern:
think this actually seems to work!

tjrileywisc:
I think Python 3 is working again but 2.7 isn't according to the last time I ran nosetests, didn't yet get a chance to run individual tests for that version. Did they complete for you according to appveyor?

PR #242, opened 2017-11-05
appveyor fix

erikbern:
think this might work


tjrileywisc:
I'm trying to get visual studio 2008 up and running to build the module there so I can get a symbol file and actually debug it when it crashes. Will update when I have more.

erikbern:
i think the fact that it runs well on python 3.6 is pretty awesome though!

should make it a bit easier to debug since it indicates it's something with python 2.7 (as opposed to the core c module)

erikbern:
let's at least start running appveyor on python 3 – great improvement

tjrileywisc:
thanks for the readme shoutout!

PR #243, opened 2017-11-06
Hamming distance (continued from #231)

erikbern:
This implements the Python binding for Hamming distance – based on #231 by @maumueller 

To make it a bit simpler:
1. I removed all the "chunking" from the C++ part. Inside the C++ module, each `uint64_t` item is considered a "dimension".
2. I kept the Python interface consistent and avoided a bunch of extra code by assuming it's passed an array of floats. This creates a few extra unnecessary copies of data and a bloated storage in Python. I think in the grand scheme of things, it's probably not substantial.
3. Removed some C++11 features, think support for things like `auto` is still a bit unreliable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,8 @@ Annoy was built by `Erik Bernhardsson <http://www.erikbern.com>`__ in a couple o
 Summary of features
 -------------------
 
-* Euclidean distance, Manhattan distance or cosine distance, where cosine distance uses Euclidean distance of normalized vectors = sqrt(2-2*cos(u, v))
+* `Euclidean distance <https://en.wikipedia.org/wiki/Euclidean_distance>`__, `Manhattan distance <https://en.wikipedia.org/wiki/Taxicab_geometry>`__, `cosine distance <https://en.wikipedia.org/wiki/Cosine_similarity>`__, or `Hamming distance <https://en.wikipedia.org/wiki/Hamming_distance>`__
+* Cosine distance is equivalent to Euclidean distance of normalized vectors = sqrt(2-2*cos(u, v))
 * Works better if you don't have too many dimensions (like <100) but seems to perform surprisingly well even up to 1,000 dimensions
 * Small memory usage
 * Lets you share memory between multiple processes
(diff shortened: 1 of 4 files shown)


erikbern:
Something is flaky with the distance calculation – seems to work most of the time but fails like 10-20% of the time. I can reproduce this behavior locally.

```
FAIL: test_basic_conversion (hamming_index_test.HammingIndexTest)
  File "/Users/erikbern/annoy/test/hamming_index_test.py", line 35, in test_basic_conversion
    self.assertAlmostEqual(i.get_distance(0, 1), numpy.dot(u - v, u - v))
  File "/Users/erikbern/annoy/test/common.py", line 20, in assertAlmostEqual
    super(TestCase, self).assertAlmostEqual(x, y, 3)
AssertionError: 46.0 != 45 within 3 places
```

erikbern:
@maumueller let me know if this is good to be merged!

erikbern:
Btw looking at the splitting code again. You're picking a random index to split on right? Seems like maybe a slightly better solution would be to sample say `sqrt(f)` items and pick the "best" split (whichever is most even). I think that's how random forests are generally built.

I might play around a bit with the algorithm to see which ones perform better.

maumueller:
> @maumueller let me know if this is good to be merged!

Removing the chunk part makes the code surprisingly clean. Looks good and thanks for your effort in including this PR!

> Btw looking at the splitting code again. You're picking a random index to split on right? Seems like maybe a slightly better solution would be to sample say sqrt(f) items and pick the "best" split (whichever is most even). I think that's how random forests are generally built.

That's a great idea. Looking at sqrt(f) coordinates should still allow for fast build times while leaving enough randomness in there. Taking a too large sample would yield too redundant  trees, I guess. I also wanted to look at more "data dependent" methods of splitting the data in the nodes.

erikbern:
> That's a great idea. Looking at sqrt(f) coordinates should still allow for fast build times while leaving enough randomness in there. Taking a too large sample would yield too redundant trees, I guess. I also wanted to look at more "data dependent" methods of splitting the data in the nodes.

Exactly – that's the tradeoff. https://en.wikipedia.org/wiki/Random_subspace_method

erikbern:
Merging this for now – will follow up at some point later and play around with other splitting methods

maumueller:
In the long run, I also want to play around with jaccard similarity. Maybe we will find some nice abstractions from there.

ymarzougui:
Thank you @erikbern and @maumueller for this PR!
Looking at the performance screenshots in #231 I see that there is a tradeoff between the number of queries per second and the recall, my question is : what n_trees  and search_k values should I have in order to ensure a maximum recall of 1?

maumueller:
Do you by any chance mean to **guarantee** recall 1? Approximate techniques will have a very hard time achieving this, and it is also the question if your application really needs "exact nearest neighbors" all the time.

Finding parameters that achieve a certain recall (say, 0.9) is usually achieved by building the data structures for smaller samples of your data set, for which you can easily build many data structures and query them. From this you build a set of candidate parameters that you might need to refine for the larger dataset, but only small changes should be necessary.  You can also fix `n_trees` (basically fixing the space you want to use) and then increase `search_k` until you achieve the recall you wish for.

ymarzougui:
I see, I did not mean exactly 1, but close to 1 as you had for the SIFT dataset, so I wondered how to achieve that the same way you had in the graphs. In my case the nearest neighbors are used for a supervised classification algorithm so I thought maximizing the recall would lead better accuracy. 
Thanks a lot for your explanation!

PR #245, opened 2017-11-20
Typo?

kant:
mmaped for memory maped or just a typo?


erikbern:
not a typo

http://man7.org/linux/man-pages/man2/mmap.2.html

PR #246, opened 2017-11-21
WIP - Build trees threaded

tjrileywisc:
Partial support for building trees threaded. It should at least compile with MSVC and gcc but currently crashes most of the time.


tjrileywisc:
Not sure what this EEE thing is about in the appveyor output. I saw this when I was building locally too:
https://ci.appveyor.com/project/erikbern/annoy/build/1.0.41#L267

thomas4g:
This definitely looks a bit nicer than my WIP! One thing I did do differently that you might consider is instead of assigning each thread a particular number of threads, I just let each thread loop with the same terminating condition - That way if one thread is a bit faster it can go ahead and start on another tree instead of just ending.

tjrileywisc:
Seems like it's failing on tests and compiles for Lua and Go.

src/annoymodule.cc
```diff
@@ -356,14 +359,15 @@ py_an_add_item(py_annoy *self, PyObject *args, PyObject* kwargs) {
 static PyObject *
 py_an_build(py_annoy *self, PyObject *args, PyObject *kwargs) {
   int q;
+  int n_threads;
   if (!self->ptr) 
     return NULL;
-  static char const * kwlist[] = {"n_trees", NULL};
-  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "i", (char**)kwlist, &q))
+  static char const * kwlist[] = {"n_trees", "n_threads", NULL};
+  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "i", (char**)kwlist, &q, &n_threads))
```
erikbern: replace `"i"` with `"i|i"` and set `int n_threads=1` as the default value. that way users don't have to provide a value (it will default to 1)

test/types_test.py
```diff
@@ -26,15 +26,15 @@ def test_numpy(self, n_points=1000, n_trees=10):
             a = a.astype(random.choice([numpy.float64, numpy.float32, numpy.uint8, numpy.int16]))
             i.add_item(j, a)
 
-        i.build(n_trees)
+        i.build(n_trees, 1)
```
erikbern: see comment higher up on default argument – should be no reason to force the user to provide this argument by default

src/annoylib.h
```diff
@@ -473,28 +475,78 @@ template<typename S, typename T, typename Distance, typename Random>
       _n_items = item + 1;
   }
 
-  void build(int q) {
+  void build_tree_threaded(int num_trees, vector<S> &_roots) {
+	  // just for this thread
```
erikbern: should use two spaces here, not tabs

erikbern: also this function shares a lot of code with `build`, would be nice to merge back

setup.py
```diff
@@ -47,9 +47,9 @@
     cputune = ['-mcpu=native',]
 
 if os.name != 'nt':
-    compile_args = ['-O3', '-ffast-math', '-fno-associative-math']
+    compile_args = ['-std=c++11','-O3', '-ffast-math', '-fno-associative-math']
```
erikbern: what's the reason we need to add support for c++11?

src/annoymodule.cc
tjrileywisc: I have ```if (!PyArg_ParseTupleAndKeywords(args, kwargs, "i|i", (char**)kwlist, &q, &n_threads))``` now is there another change needed for that line?

There should be a default for n_threads in annoylib.h, does that work?
```void build(int q, int n_threads = 1)```

File: test/types_test.py (hunk 2 above)
Comment (tjrileywisc) [reply test/types_test.py]: Should be fixed now

File: src/annoymodule.cc (hunk 1 above)
Comment (tjrileywisc) [reply src/annoymodule.cc]: Fixed now (I think)

File: setup.py (hunk 4 above)
Comment (tjrileywisc) [reply setup.py]: I got a lot of 'thread' is not a member of std and similar warnings for other <thread> functions. Would compiling for c++ 11 cause issues for *nix users?

Comment (thomas4g) [reply setup.py]: This is why I used pthreads in my branch, but if that doesn't work on Windows... is that a lose-lose then? What's the de-facto cross-platform threading?

**Comment by thomas4g** [comment]
@tjrileywisc Yeah - if you look at my branch, I used a separate counter to track `_roots.size()` and used the gcc atomic builtins to atomically fetch/increment.

File: src/annoylib.h [hunk 5]
```diff
@@ -473,28 +475,78 @@ template<typename S, typename T, typename Distance, typename Random>
       _n_items = item + 1;
   }
 
-  void build(int q) {
+  void build_tree_threaded(int num_trees, vector<S> &_roots) {
```
erikbern: should pass mutable datastructures as pointers, not references, i.e. `vector<S>* roots`

setup.py
a1k0n: You could use boost::thread, maybe, since it's already using boost python?

src/annoylib.h
```diff
@@ -473,28 +475,78 @@ template<typename S, typename T, typename Distance, typename Random>
...
+		  vector<S> indices;
+		  for (S i = 0; i < _n_items; i++) {
+			  if (_get(i)->n_descendants >= 1) // Issue #223
+				  indices.push_back(i);
+		  }
+		  S tree = _make_tree(indices);
+		  _roots_thread.push_back(tree);
+	  }
+	  // push to our shared roots vector
+	  _roots.insert(_roots.end(), _roots_thread.begin(), _roots_thread.end());
```
a1k0n: this needs to be protected by a mutex.

tjrileywisc: It shouldn't be using boost python (I think an earlier version of annoy did though).

a1k0n: Oh, that's right... I forgot. Hm.

setup.py
erikbern: annoy doesn't use boost any more since it caused a lot of problems installing annoy, so i'm not too excited about reintroducing it

erikbern: could maybe just have some `#ifdef` magic and only support threading on posix systems?

src/annoylib.h
erikbern: seems easier if each thread just returns a vector of roots and the main thread pushes them into the roots array

setup.py
tjrileywisc: Is OpenMP an option?

erikbern: would avoid anything that require any linking with libraries just because it gets more complex to install

a1k0n: well, then stick with c++11 then, since std::thread was standardized in c++11?

erikbern: ok seems like that's the logical conclusion :)

tjrileywisc:
Okay, I think I've got all of the changes above integrated but there's some cleanup to do.

Good news- seems to compile everywhere and is not crashing. I did see a speedup when building the precision_test.cpp example and generating 10,000 vectors and building the index.

tjrileywisc:
Tests seem to be failing without clear error messsages.
I know why the Lua/Go tests would fail (I don't know how to tell them to compile with the -std=c++11 option) but not sure why the python tests are failing.

src/annoylib.h
```diff
@@ -84,6 +85,10 @@ using std::string;
 using std::pair;
 using std::numeric_limits;
 using std::make_pair;
+using std::mutex;
+
+mutex _node_mutex;
+mutex _roots_mutex;
```
erikbern: no reason to make these global

tjrileywisc: I was trying to put them in the class with everything else, but I kept getting 'deleted function' errors when compiling.

tjrileywisc: The exact warning (when compiling the precision_test.cpp example in Linux with the s_compile_cpp.sh script):

`error: use of deleted function ‘std::mutex::mutex(const std::mutex&)`

(similar warnings with MSVC).

erikbern: strange! i can fix later

tjrileywisc:
I also added a compiler option for MSVC for floating point in setup.py ('-ffast-math' equivalent I think). It's incidental to the CI errors though (probably).

erikbern:
nice to see more activity on this one – let me know if there's any way for me to help!

tjrileywisc:
Yeah, I took a break on this one. It's flaky at the moment, usually crashing Python on Windows, or running and failing tests (not consistently). The precision_test.cpp example works fine in linux, but then I can't get the python nosetests to run there. It's almost as if nose isn't finding the tests at all (I tried `nosetests -v` without any difference in output).

tjrileywisc:
It looks as if python is crashing when running the accuracy tests (it failed right away on the fashion mnist dataset), right at this line:

`js_fast = annoy.get_nns_by_vector(v, 10, 1000)`

I added something similar to the precision_test.cpp example (albeit not with this dataset):

```
double *vec2 = (double *)malloc(f * sizeof(double));

for (int z = 0; z<f; ++z) {
	vec2[z] = (distribution(generator));
}
t.get_nns_by_vector(vec2, K, n, &closest, nullptr);
closest.clear(); vector<int>().swap(closest);
```
... which didn't crash, so I guess that means the python module code for this function has an issue with my threading changes.

erikbern:
Any update on this?

tjrileywisc:
I'm having trouble figuring out why accuracy tests are failing... I'm going to close this PR for now since I think I need to learn a lot more C++ before I can get this.

erikbern:
Ok, I’ll take a look to see what can be done

erikbern:
looks like @tjrileywisc gave up sadly, but feel free to give it a shot!

PR #247, opened 2017-11-24
Run accuracy test using pre-built vectors

erikbern:
should speed up tests a lot


tjrileywisc:
Looks like we need a `pip install h5py` in the appveyor script.

PR #248, opened 2017-11-27
fix holes test

yonromai:


erikbern:
what was broken with the previous code? it would add vectors at every [k * sqrt(2)] position, so basically leave random gaps

```
>>> [int(i*2**0.5) for i in range(40)]
[0, 1, 2, 4, 5, 7, 8, 9, 11, 12, 14, 15, 16, 18, 19, 21, 22, 24, 25, 26, 28, 29, 31, 32, 33, 35, 36, 38, 39, 41, 42, 43, 45, 46, 48, 49, 50, 52, 53, 55]
```

erikbern:
maybe more instructive would be to do something like `random.choice(range(1000), 500)`

yonromai:
Your code was `i*2**-0.5`:
```python
t = set([int(i*2**-0.5) for i in range(1000)])
max(t) # 706
len(t) # 707
```

erikbern:
haha ok that's not ideal

test/holes_test.py
```diff
@@ -30,12 +30,10 @@ def test_holes(self):
     def test_holes_more(self):
         f = 10
         index = AnnoyIndex(f)
-        valid_indices = set()
-        for i in range(1000):
-            i2 = int(i*2**-0.5) # leave holes every few items
-            valid_indices.add(i2)
+        valid_indices = range(0, 2000, 2) # leave holes
```
erikbern: do you mind replacing this with `random.sample(range(2000), 1000)` to make it a bit uneven?

erikbern:
LGTM but would have a minor preference for a more random hole distribution

yonromai:
I bumped into this because I think the code with holes doesn't work here:
https://github.com/spotify/annoy/blob/master/src/annoymodule.cc#L233
It crashes when you have holes because n_items != max(items)

test/holes_test.py
yonromai: sure thing

yonromai:
maybe not, looks like the tests are still passing. The problem is probably in my code :)

erikbern:
yeah i think n_items() will just return max(items) + 1 so shouldn't be an issue

erikbern:
thanks @yonromai 

are you on the latest version of annoy yet? would love to hear if it leads to any improvements!

yonromai:
yes, looking into it right now :)

PR #255, opened 2017-12-17
silence some g++ complaints

eddelbuettel:
This silences a few nags I got from g++ 6.3 when updating to your most recent sources and enabling Hamming.  One or two were obvious signed/unsigned comparison, and I also `#ifdef`-ed away the pragam. Nothing major here, passed all Travis tests and should not cause harm.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -49,8 +49,10 @@ typedef signed __int32    int32_t;
 #include <queue>
 #include <limits>
 
+#ifdef _MSC_VER
 // Needed for Visual Studio to disable runtime checks for mempcy
 #pragma runtime_checks("s", off)
+#endif
 
 // This allows others to supply their own logger / error printer without
 // requiring Annoy to import their headers. See RcppAnnoy for a use case.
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Hm, can't speak to AppVeyor blowing up.  

Travis passed for [me on all subtests, including Go](https://travis-ci.org/eddelbuettel/annoy/builds/317533134).  Any idea why swig all of a sudden goes funky on you?

erikbern:
might just be flakiness, let's try to restart the build. weird that both appveyor and travis flaked out

eddelbuettel:
Yay for random flakyness. Now we're 2 for 2.

src/annoylib.h
```diff
@@ -230,7 +232,7 @@ struct Hamming {
 
   template<typename T>
   static inline T pq_distance(T distance, T margin, int child_nr) {
-    return distance - (margin != child_nr);
+    return distance - (margin != (unsigned int) child_nr);
```
erikbern: think this could be just `(T)child_nr` but not blocking

erikbern:
thanks @eddelbuettel

src/annoylib.h
eddelbuettel: That's a good one.

eddelbuettel: Tested, and now committed, here.

erikbern: i'm planning to change the splitting code for Hamming distance anyway, so not important

PR #262, opened 2018-01-13
Manually vectorize distance functions to improve performance

ReneHollander:
For a start I vectorized the euclidian distance with AVX intrinsics. I wanted to make sure, that there is an actual improvement. I created a super [basic benchmark](https://gist.github.com/ReneHollander/4ca884ceb1537f013aa8d0f868e9b844) which improved the index build time with `2*f` trees from roughly 45 seconds to 35 seconds and search time from 13ms to 9ms (Windows 10 Subsystem for Linux on an i7 4770K overclocked to 4GHz, the benchmark also contains some results). I only tested with random high dimensional data (96 and 100 dimensions and 100000 features)

With that said, is there interest in optimizing other distance functions and other functions in general? (I am no expert and there is the possibility, that there are even better solutions, but for a start it could help)

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -70,6 +70,18 @@ typedef signed __int32    int32_t;
 #endif
 
 
+#if !defined(NO_MANUAL_VECTORIZATION)
+#if defined(__AVX__) && defined (__SSE__) && defined(__SSE2__) && defined(__SSE3__)
+#define USE_AVX
+#elif defined (__SSE__) && defined(__SSE2__) && defined(__SSE3__)
+#define USE_SSE
+#endif
+#endif
+
+#if defined(USE_AVX) || defined(USE_SSE)
+#include <x86intrin.h>
+#endif
+
 #ifndef ANNOY_NODE_ATTRIBUTE
     #ifndef _MSC_VER
         #define ANNOY_NODE_ATTRIBUTE __attribute__((__packed__))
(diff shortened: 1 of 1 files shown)


erikbern:
that's pretty cool! I think the small differences in the distances is totally acceptable (after all, the A in Annoy stands for Approximate)

I might run some tests and merge this if I can reproduce the speedup!

a1k0n:
Would be good to try it on the dot product (for cosine distance, and for every single plane test when traversing the tree) also, if it's not already auto-vectorized...

erikbern:
Good point @a1k0n – should add that too and test!

ReneHollander:
Sure! I will try to vectorize some more functions (first with AVX) and later with SSE. Sadly I am not able to look into AVX512, because my CPU doesn't support it.

ReneHollander:
After some testing I found an issue in the way features with a dimension that is not a multiple of 8 are handled. After fixing those calculations, all tests pass now.
I also optimized the dot product in the margin function and the angular distance. Here are the results: https://gist.github.com/ReneHollander/4def83f7755c6d63e1efb120774f2306
The speedups for index building are quite dramatic, but also the query time got reduced quite a bit.
I did some profiling with valgrind and also found a hotspot with the inner loops in `two_means`. I don't think optimizing it will improve the speed as much as my other changes did, but it could help. Some goes with the vector norm.

ReneHollander:
I also forgot to add, that those optimizations are enabled if AVX support is found (building with `-march=native`, as it is currently done in `setup.py`, will pick it up). One can override the hand vectorization with defining `NO_MANUAL_VECTORIZATION` before including `annoylib.h`.

I also don't know if it is of use adding support for only SSE CPUs. Every modern CPU since 2011 has support for AVX.

a1k0n:
:+1: amazing! Nice work!

erikbern:
excellent – will take a look at this in the next few days

ReneHollander:
`get_norm`, `normalize` and `two_means` are now also optimized with AVX intrinsics. This only speeds up trees using angular distance ([Benchmark](https://gist.github.com/ReneHollander/98f8a231fc1725fc2e3e388bff5ed741), baseline is with hand optimized distance and margin), euclidian distance is completly unaffected, as expected.

There are also intrinsics for popcount, so it should be possible to optimize hamming distance by hand as well. Also manhattan distance is missing the optimizations for `distance` (I have to look into the absolute value calculation using AVX first).

ReneHollander:
I decided against vectorizing hamming distance calculation. Because the data is packed into 64bit integers, it would only make sense when you have features with over 256 dimensions.
Other than that, everything is finished and it just needs to be reviewed 😄

erikbern:
This is great – let me take a look soon!

erikbern:
Thanks @ReneHollander – this is great!

I think the code is a bit repetitive so let me refactor and share a PR shortly!

src/annoylib.h
```diff
@@ -69,6 +69,36 @@ typedef signed __int32    int32_t;
...
+// Horizontal single sum of 256bit vector.
+static inline float hsum256_ps_avx(__m256 v) {
+  const __m128 x128 = _mm_add_ps(_mm256_extractf128_ps(v, 1), _mm256_castps256_ps128(v));
+  const __m128 x64 = _mm_add_ps(x128, _mm_movehl_ps(x128, x128));
+  const __m128 x32 = _mm_add_ss(x64, _mm_shuffle_ps(x64, x64, 0x55));
+  return _mm_cvtss_f32(x32);
+}
+#endif
+
+#ifdef USE_SSE
```
erikbern: actually this function doesn't seem to be used?

src/annoylib.h
```diff
@@ -137,6 +167,108 @@ inline void two_means(const vector<Node*>& nodes, int f, Random& random, bool co
...
+    v += 8;
+  }
+  // Don't forget the remaining values.
+  for (; i > 0; i--) {
+    *v /= norm;
+    v++;
+  }
+}
+
+template<typename T, typename Random, typename Distance, typename Node>
```
erikbern: why is this functino actually needed? Shouldn't the distance function already use AVX when possible?

src/annoylib.h
ReneHollander: It is currently not needed.

src/annoylib.h
ReneHollander: The distance function will use the optimized code, but I also did some optimizations in the inner for loops. Without them I was not able to measure any improvements when normalize and get_norm where also optimized.

ReneHollander:
You are right, the code is pretty repetitive. Maybe it is possible to move out the dot product calculation. I also fixed some potential windows incompatibility (and the unused code) in another pull request: #266

erikbern:
@ReneHollander did you see #265 ? I factored out two helpers

* `dot(p, q)` for dot product
* `dot3(p, q)` for computing the quantities pq, pp, and qq

PR #265, opened 2018-01-19
Simplify AVX usage

erikbern:
Follow up on @ReneHollander's PR #262 

Took a quick stab at simplifying AVX usage. Almost all of it was various types of dot products, so I added a function `template<T> dot(const T* x, const T* y, int f)` – if `USE_AVX` is defined then I have a template specialization for T = float that implements it using AVX

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -99,8 +99,10 @@ using std::pair;
 using std::numeric_limits;
 using std::make_pair;
 
+namespace {
+
 template<typename T>
-static inline T get_norm(T* v, int f) {
+inline T get_norm(T* v, int f) {
   T sq_norm = 0;
   for (int z = 0; z < f; z++)
     sq_norm += v[z] * v[z];
@@ -108,7 +110,7 @@ static inline T get_norm(T* v, int f) {
 }
 
 template<typename T>
-static inline void normalize(T* v, int f) {
+inline void normalize(T* v, int f) {
   T norm = get_norm(v, f);
   for (int z = 0; z < f; z++)
     v[z] /= norm;
@@ -143,7 +145,7 @@ inline void dot3(const T* x, const T* y, int f, T* xx, T* yy, T* xy) {
 
 #ifdef USE_AVX
 // Horizontal single sum of 256bit vector.
-static inline float hsum256_ps_avx(__m256 v) {
+inline float hsum256_ps_avx(__m256 v) {
   const __m128 x128 = _mm_add_ps(_mm256_extractf128_ps(v, 1), _mm256_castps256_ps128(v));
   const __m128 x64 = _mm_add_ps(x128, _mm_movehl_ps(x128, x128));
   const __m128 x32 = _mm_add_ss(x64, _mm_shuffle_ps(x64, x64, 0x55));
@@ -204,7 +206,7 @@ inline void dot3<float>(const float* x, const float *y, int f, float* xx, float*
 }
 
 template<>
-static inline float get_norm<float>(float *v, int f) {
+inline float get_norm<float>(float *v, int f) {
   float sq_norm = 0;
   int i = f;
   if (f > 7) {
@@ -226,7 +228,7 @@ static inline float get_norm<float>(float *v, int f) {
 }
 
 template<>
-static inline void normalize<float>(float *v, int f) {
+inline void normalize<float>(float *v, int f) {
   float norm = get_norm(v, f);
   __m256 v_norm = _mm256_set1_ps(norm);
 
@@ -244,7 +246,7 @@ static inline void normalize<float>(float *v, int f) {
 #endif
 
 template<typename T, typename Random, typename Distance, typename Node>
-static inline void two_means(const vector<Node*>& nodes, int f, Random& random, bool cosine, T* iv, T* jv) {
+inline void two_means(const vector<Node*>& nodes, int f, Random& random, bool cosine, T* iv, T* jv) {
   /*
     This algorithm is a huge heuristic. Empirically it works really well, but I
     can't motivate it well. The basic idea is to keep two centroids and assign
@@ -278,6 +280,8 @@ static inline void two_means(const vector<Node*>& nodes, int f, Random& random, 
     }
   }
 }
+
+} // namespace
 
 struct Angular {
   template<typename S, typename T>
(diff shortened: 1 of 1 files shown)


src/annoylib.h
```diff
@@ -296,81 +252,15 @@ struct Angular {
     // want to calculate (a/|a| - b/|b|)^2
     // = a^2 / a^2 + b^2 / b^2 - 2ab/|a||b|
     // = 2 - 2cos
-    T pp = 0, qq = 0, pq = 0;
-    for (int z = 0; z < f; z++, x++, y++) {
-      pp += (*x) * (*x);
-      qq += (*y) * (*y);
-      pq += (*x) * (*y);
-    }
+    T pp = dot(x, x, f), qq = dot(y, y, f), pq = dot(x, y, f);
```
erikbern: I wonder if rewriting it this way will be slower because of cache locality?

src/annoylib.h
```diff
@@ -204,70 +224,6 @@ static inline void normalize(float *v, int f) {
     v++;
   }
 }
-
-template<typename T, typename Random, typename Distance, typename Node>
```
erikbern: actually realized the AVX version I just removed had some speedups that were unique to it - when it updates `iv` and `jv` it actually won't benefit from AVX in the current version. I'm tempted to make that sacrifice since tree building isn't usually a big issue (Annoy is already very fast) and I found the AVX version to be much harder to read (and creates a bit of code redundancy)

src/annoylib.h
```diff
@@ -554,36 +416,8 @@ struct Minkowski {
...
-    }
-    // Don't forget the remaining values.
-    for (; i > 0; i--) {
-      const float t = *x - *y;
-      result += t * t;
-      x++;
-      y++;
-    }
-    return result;
+    return dot(x, x, f) + dot(y, y, f) - 2 * dot(x, y, f);
```
erikbern: also possibly not quite as fast because of cache locality? feels like @a1k0n would know

a1k0n: Yeah, if the various components of `x` and `y` are already loaded in registers, seems like you might want to roll them up while you have them.

Run a benchmark and see!

src/annoylib.h
erikbern: hm yeah actually now that I'm looking at the AVX code, it definitely benefits from doing all of those at the same time (can load into a vector only twice, not six times)

let me add another specialized function that computes the dot products u*u, v*v and u*v at the same time – would be faster

src/annoylib.h
erikbern: already have a fix coming up, 1 sec

erikbern:
Ok I added a new primitive dot3 to compute the products xx, yy, and xy at the same time now. Think it should do the trick!

erikbern:
Unscientific benchmark:

After this PR:
Building + searching: 246.822s
Searching (best of 3): 8.584s

Current master:
Building + searching: 232.608s
Searching (best of 3): 8.243s

Master from before the AVX instructions
Building + searching: 230.247s
Searching (best of 3): 8.505s

This is really weird. No difference? I know for a fact that my laptop uses AVX. Maybe it doesn't  give any speedup on my CPU (?)

erikbern:
seems like my processor might just be too slow: https://stackoverflow.com/questions/43245787/vectorization-speed-up-expected-for-sse-avx-and-avx2

i'll try on a different CPU tomorrow

erikbern:
rebased on master (#266 in particular)

ReneHollander:
I checked the assembly and without declaring `get_norm` and `normalize` like you did with `dot3` and `dot`, g++ won't use it. The float implementation of `two_means` forced the usage of the float variants and thus created the speedup. My optimized `two_means` is mostly unreadable and does not affect performance at all.
```diff
Index: src/annoylib.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/annoylib.h	(revision 35be9a1b8ff8c0f7ccfece330b1386691993c0ae)
+++ src/annoylib.h	(date 1516371870000)
@@ -237,7 +237,8 @@
   }
 }
 
-static inline float get_norm(float *v, int f) {
+template<>
+inline float get_norm<float>(float *v, int f) {
   float sq_norm = 0;
   int i = f;
   if (f > 7) {
@@ -258,7 +259,8 @@
   return sqrt(sq_norm);
 }
 
-static inline void normalize(float *v, int f) {
+template<>
+inline void normalize<float>(float *v, int f) {
   float norm = get_norm(v, f);
   __m256 v_norm = _mm256_set1_ps(norm);
 

```
I also ran my simple benchmarks, and I still see the same results I posted in my other PR. Index build time gets halfed, Query time reduced by 25% or something like that.

erikbern:
Ok great – I'm glad my rusty C++ intuition was correct and template specialization is the way to go here.

I'll probably merge this for now

ReneHollander:
I did some more benchmarking (all synthetic data):
There is a noticable improvement on my i7 4770K in WSL on Windows 10. At no point is the AVX optimized code slower, and it performs the best if the number if dimensions is a multiple of 8 (as expected). I think I set the k for the search test a bit too low, but I did not want to rerun the test (I could not use my PC for 6 hours...). I hope you see similar improvements in your tests!
# Euclidian
euclidian_build_time
euclidian_search_time
(Search time in us)
# Angular
angular_build_time
angular_search_time

erikbern:
Awesome! Was actually just running some benchmarks myself, this time on a c4.2xlarge:

current master:
glove-25 build+search: 136.447s
glove-25 search (best of 3): 5.697s
fashion-mnist build+search: 19.126s
fashion-mnist search (best of 3): 8.798s

pre-avx:
glove-25 build+search: 152.749s
glove-25 search (best of 3): 6.351s
fashion-mnist build+search: 33.285s
fashion-mnist search (best of 3): 18.096s

this branch:
glove-25 build+search: 146.448s
glove-25 search (best of 3): 5.754s
fashion-mnist build+search: 20.259s
fashion-mnist search (best of 3): 9.163s

looks like master and this branch is basically same (great, my refactoring didn't ruin anything) and much faster than pre-avx, by up to a 2x factor. very nice @ReneHollander !

PR #274, opened 2018-02-01
On disk index build

### LINKED ISSUE #85: When the data size is too large to fit into the memory?
I got 50 million 1000 dim vectors need too be indexed, but I don't get that much ram to hold all the data during building the tree. Is possible to modify annoy to support that amount of data? Any hint on how to do that?

### LINKED ISSUE #250: Python crashes when loading large index on windows
Python crashes when loading large index (both 10g one and a 20g one), but works fine with a 100m test index.

Environment: 
windows server 2016
miniconda python 3.6
annoy 1.10.0
vcbuildtools 2015

ReneHollander:
Index build can now be performed directly on disk. The resulting index does not need to fit into memory anymore.
This fixes #85 and probably fixes #250.

TODO:
- [x] Add support for `on_disk_build` to LUA module
- [x] Add support for `on_disk_build` to Go module
- [x] Fix Windows support (there is not `ftruncate` on windows)
- [ ] Fix WSL compatability (`ftruncate` to cut off extra allocated memory fails)
- [ ] Check return value of `ftruncate`


erikbern:
cool!

erikbern:
does this really help conserve memory? you still need to map the whole file into primary memory at some point or else querying will be prohibitively slow (on the order of 100-1000x slower than RAM)

ReneHollander:
I did not do any speed comparisons yet, but my little test program runs only a bit slower. I think the OS is doing a good job caching the pages. (I used an SSD to store the index)

erikbern:
Sure, I think with an SSD, the delta between cached and not cached (on disk) is about 10-30x still though. But I guess if speed is not a major concern, it's nice to be able to build indexes that are orders of magnitude larger

erikbern:
I think this PR will make it a lot more stable when building an index that's within [x/2, x] where x is the total RAM on your machine. So that seems like a big benefit

erikbern:
Hey what's the status of this?

ReneHollander:
I did not get around to test it on native linux yet (my arch install broke some time ago).
But if you are still interested in including it, I will put some more time into it and make sure it is tested on all platforms :)

erikbern:
Ok, cool, let me take a look at this and get back to you!

AjayP13:
@ReneHollander this looks exactly like what I was looking for. Any updates on finishing this PR up? I would like to use this feature. Would be happy to help if needed, I think I understand most of these changes.

ReneHollander:
I am still interested in finishing this PR and doing proper performance analysis. I still think it could be useful, especially with the recent developments off SSD technology (eg Intel Optane).
I think I could work on it some more next week.

erikbern:
nice, let me know if i can help!

ReneHollander:
The only question is if you would want something like this to be included in annoy. If it is something you are not so sure about, I would rather not invest the time. I don't really need support for this usecase in my project at the time of writing this, but I would be happy working on it.

erikbern:
To be honest it's probably not a game changer, more a nice to have. Hope that's helpful.

ReneHollander:
I also get your concerns regarding memory locality, but you should at least be able to build an index where each tree fits into memory in a reasonable amount of time. As only one tree is built at a time, the OS should keep the needed pages in memory, while keeping trees already built on disk. When building 20 trees, it could dramatically increase the amount of items the index can hold if you are bound by RAM.

AjayP13:
I think it would allow creating a large index on a machine bounded by low
RAM (development machine) for use on a larger machine with sufficient RAM
(prod machine). Currently, this is not possible. That's my use case at
least.

szkocot:
This PR helped me a lot in my recent project, where I would need 100-200GB or more of RAM to index whole data with annoy. With SSD I was able to index whole data with very little performance hit. SSDs are getting cheap (130-150 USD/TB) and it would be a great feature to have in annoy officially.

The only drawback is It didn't work (with very much of rows) under Windows 10. I was getting a length of indexed annoy model == -1 (like others reported), but this isn't something very important as Linux is better for such applications anyway.

ReneHollander:
The build seems to still be passing after merging with master. Maybe @erikbern can take a look at the API and if it should be done differently. After that I will do some more cleanup of the code related to memory management.

erikbern:
This looks good – let me review a bit more thoroughly in the next week or so. My 2nd child was just born last week so I'm a bit behind!

ArielSSchwartz:
This is a very useful feature. +1 for merging into the official release.

erikbern:
forgot about this but will review shortly!

erikbern:
@ReneHollander this looks good! Some thoughts

1. I'd like a few more unit tests. Will add myself.
2. I think (like you said) the memory management code is a bit messy so I'd appreciate post-merge cleanup
3. I think the API is a bit confusing. Thinking about it more I wonder if it really should just be a part of the constructor of `AnnoyIndex` where either it takes a filename (in which case it builds on disk) or not (in which case it builds in-memory). I think in fact at that point you could even deprecate the `save` method and remove it in a future version. What do you think? I don't see a reason why you'd want to build it in memory and then save it to disk rather than just build it onto disk from scratch.

erikbern:
thanks for the contribution @ReneHollander !

ReneHollander:
Thanks for merging! I will go over the memory management first (maybe I can get rid of some [unnecessary mallocs](https://github.com/spotify/annoy/blob/master/src/annoylib.h#L371)).

If we change the API like that, I guess it would cause problems for people using the library. It is a pretty big change. Are there any other things with the API that you think need to be looked at? Then we could do it in a bigger release (maybe 2.0 if we consider semver?).

erikbern:
@ReneHollander i don't think it necessarily has to be that problematic if there's enough of a deprecation period of the `save` method (like, six months maybe?)

i don't have super strong feelings though. let me know what you think.

erikbern:
(and yeah, 2.0 would be one way to justify backwards incompatibility)

src/annoylib.h
```diff
@@ -851,18 +894,23 @@ template<typename S, typename T, typename Distance, typename Random>
     _n_items = 0;
     _n_nodes = 0;
     _nodes_size = 0;
+    _on_disk = false;
     _roots.clear();
   }
 
   void unload() {
```
louisabraham, line 901: Shouldn't one call [msync](https://man7.org/linux/man-pages/man2/msync.2.html) before munmap to ensure the modifications are flushed?

disclaimer: I'm reading this code to learn about mmap.

src/annoylib.h
```diff
@@ -851,18 +894,23 @@ template<typename S, typename T, typename Distance, typename Random>
...
-    if (_fd) {
-      // we have mmapped data
+    if (_on_disk && _fd) {
       close(_fd);
```
louisabraham, line 903: Shouldn't one call [msync](https://man7.org/linux/man-pages/man2/msync.2.html) before munmap to ensure the modifications are flushed?

disclaimer: I'm reading this code to learn about mmap.

src/annoylib.h
```diff
@@ -782,7 +798,24 @@ template<typename S, typename T, typename Distance, typename Random>
...
+  bool on_disk_build(const char* file) {
+    _on_disk = true;
+    _fd = open(file, O_RDWR | O_CREAT | O_TRUNC, (int) 0600);
+    if (_fd == -1) {
+      _fd = 0;
+      return false;
+    }
+    _nodes_size = 1;
+    ftruncate(_fd, _s * _nodes_size);
+#ifdef MAP_POPULATE
```
louisabraham, line 811: I think one could test using [madvise](https://man7.org/linux/man-pages/man2/madvise.2.html) with MADV_RANDOM to improve performance: it prevents read-ahead by telling the kernel to expect random access.

src/annoylib.h
louisabraham: I think it should be fine: https://linux.die.net/man/2/munmap

>     Share this mapping. Updates to the mapping are visible to other processes that map this file, and are carried through to the underlying file. The file may not actually be updated until msync(2) or munmap() is called.

PR #276, opened 2018-02-05
fix issue with holes in the index and angular distance

### LINKED ISSUE #273: Annoy has terrible performance on the nytimes-256-angular dataset
See https://github.com/erikbern/ann-benchmarks

Noticed Annoy really sucks on this dataset. My first theory was it had something with swapping to do but no, the third run isn't faster than the first.

My other theory is that it has something to do with how this dataset is "skewed" since it's built by taking a PCA of a sparse matrix. That means the first dimension has a lot more variance than the second etc...

image

erikbern:
Fixes #273 

Btw @yonromai didn't you mention you had issues with "holes" in the index? This bugfix should dramatically improve performance for such indexes


erikbern:
Will push a new version to pypi tonight

PR #281, opened 2018-03-19
Fixes to Hamming distance.

maumueller:
This fixes some issues that have been introduced during #243. First, since `f` is in `uint64_t` chunks, we have to manually compute the dimension of the vectors to choose a correct index for splitting. Next, the `pq_distance` computation assumed integer, not unsigned integer. Changed it to work for unsigned integers.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -336,12 +336,12 @@ struct Hamming {
 
   template<typename T>
   static inline T pq_distance(T distance, T margin, int child_nr) {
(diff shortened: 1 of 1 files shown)


erikbern:
nice!

erikbern:
Btw added a note here – we should resurrect the Hamming tests: https://github.com/erikbern/ann-benchmarks/issues/68

erikbern:
Forgot to merge this, sorry

Issue #44, opened 2015-03-30
Add Maximum Inner Product Search


a1k0n:
See http://arxiv.org/pdf/1410.5410v2.pdf

This is pretty easy to implement. I will probably do it myself, but in case I don't get to it, I wanted to jot some notes down and this seems like a good place to do it.

The basic algorithm is very simple and can be used in Annoy Angular mode with the following modifications, assuming we use the recommended m=2, U=0.75 parameters in the paper:

Initialization:
- Add two to the input dimension before any add_item has been called, and pad the last two dimensions with 0s in add_item.

When it is time to `build()`:
- First, find the norm of the largest vector in the input set and compute a scale parameter U / ||x_max||, and then re-scale all vectors by this factor. Re-scaling the universe does not change the argmax of any maximum inner product.
- While rescaling, set the extra two dimensions of each vector to [0.5 - ||x_i||^2, 0.5 - ||x_i||^4]
- Resume the standard build process.

When it is time to query:
- Just append two zeros to the query vector. That's it. The result will be the maximum dot product, instead of the maximum cosine distance, because of the brilliant / insane proofs in the paper above.

a1k0n:
Hmm, and last modification: re-sort based on inner products instead of cosine distance.

a1k0n:
Hmm, and actually the asymmetric LSH stuff doesn't even have to affect the query algorithm at all, only the tree building phase, and so it actually only affects the way the tree nodes are split; the extra two dimensions don't need to be stored.

erikbern:
Yes and it sounds pretty easy to construct a hyperplane if you sample two point from the distribution. If you sample p and q then the boundary becomes `dot(p, x) = dot(q, x)` which is equivalent to `dot(p-q, x)=0` which means you can just compute the sign of `p-q` and `x` for any vector. The problem with this idea though is that if `|p| >> |q|` then almost all vectors will be on the same side of the split. I don't know if that's an issue in practice

a1k0n:
Well, you have to keep in mind that the magnitudes are all scaled to be < 0.75 first, so the last two components are < 0.5 and > -0.06 or so.

a1k0n:
...but then those buckets would be useless on the query side, where those coordinates are 0. Hmm.

erikbern:
I meant without adding extra coordinates, just doing the split in the original space

nemo83:
HI, was this ever implemented?

erikbern:
nope

[...]

PR #303, opened 2018-06-27
Add dot product as a distance metric.

psobot:
This PR:
 - adds a `DotProduct` metric, using the negative of the dot product between two vectors as a pseudo-distance metric.
 - makes this metric available under the string name `"dot"` in the standard Annoy interface
 - adds tests for this new metric, similar to the existing tests for the `Euclidean` metric.

You might ask - why add `dot` when we already have `angular`, which is basically just `dot` but divided by magnitude? Well, for some recommender systems (including some matrix factorization collaborative filters), the dot product between the user and item pair is a prediction of affinity between that user and item. ([Here's a great example](https://github.com/jfkirk/tensorrec/blob/master/tensorrec/tensorrec.py#L30) from @jfkirk.)

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -416,7 +416,7 @@ struct DotProduct : Angular {
 
   template<typename S, typename T>
   static inline T margin(const Node<S, T>* n, const T* y, int f) {
-    return dot(n->v, y, f);
+    return dot(n->v, y, f) + (n->dot_factor * n->dot_factor);
   }
 
   template<typename S, typename T, typename Random>
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -101,6 +101,11 @@ using std::make_pair;
 
 namespace {
 
+template<typename S, typename Node>
+inline Node* get_node_ptr(const void* _nodes, const size_t _s, const S i) {
+  return (Node*)((uint8_t *)_nodes + (_s * i));
+}
+
 template<typename T>
 inline T dot(const T* x, const T* y, int f) {
   T s = 0;
@@ -210,6 +215,7 @@ inline void two_means(const vector<Node*>& nodes, int f, Random& random, bool co
   j += (j >= i); // ensure that i != j
   memcpy(p->v, nodes[i]->v, f * sizeof(T));
   memcpy(q->v, nodes[j]->v, f * sizeof(T));
+
   if (cosine) { normalize(p->v, f); normalize(q->v, f); }
   Distance::init_node(p, f);
   Distance::init_node(q, f);
@@ -236,10 +242,20 @@ inline void two_means(const vector<Node*>& nodes, int f, Random& random, bool co
     }
   }
 }
-
 } // namespace
 
-struct Angular {
+struct Base {
+  static inline int internal_dimensions() {
+    return 0;
+  }
+
+  template<typename T, typename S, typename Node>
+  static inline void preprocess(void* nodes, size_t _s, const S node_count, const int f) {
+
+  }
+};
+
+struct Angular : Base {
   template<typename S, typename T>
   struct ANNOY_NODE_ATTRIBUTE Node {
     /*
@@ -324,7 +340,65 @@ struct Angular {
   }
 };
 
-struct Hamming {
+
+struct DotProduct : Angular {
+  static const char* name() {
+    return "dot";
+  }
+
+  static inline int internal_dimensions() {
+    return 1;
+  }
+
+  template<typename S, typename T>
+  static inline T distance(const Node<S, T>* x, const Node<S, T>* y, int f) {
+    return -dot(x->v, y->v, f);
+  }
+
+  template<typename T>
+  static inline T normalized_distance(T distance) {
+    return distance;
+  }
+
+  template<typename T, typename S, typename Node>
+  static inline void preprocess(void* nodes, size_t _s, const S node_count, const int f) {
+    // Step one: compute the norm of each vector and store that in its extra dimension (f-1)
+    for (S i = 0; i < node_count; i++) {
+      Node* node = get_node_ptr<S, Node>(nodes, _s, i);
+      T norm = sqrt(dot(node->v, node->v, f - 1));
+      if (isnan(norm)) norm = 0;
+      node->v[f - 1] = norm;
+    }
+
+    // Step two: find the maximum norm
+    T max_norm = 0;
+    for (S i = 0; i < node_count; i++) {
+      Node* node = get_node_ptr<S, Node>(nodes, _s, i);
+      if (node->v[f - 1] > max_norm) {
+        max_norm = node->v[f - 1];
+      }
+    }
+
+    // Step three: set each vector's extra dimension to sqrt(max_norm^2 - norm^2)
+    for (S i = 0; i < node_count; i++) {
+      Node* node = get_node_ptr<S, Node>(nodes, _s, i);
+      T node_norm = node->v[f - 1];
+
+      T extra_dimension = sqrt(pow(max_norm, 2.0) - pow(node_norm, 2.0));
+      if (isnan(extra_dimension)) extra_dimension = 0;
+
+      node->v[f - 1] = extra_dimension;
+
+      Angular::init_node(node, f);
+    }
+  }
+
+  static inline int default_search_k_multiplier() {
+    return 50;
+  }
+};
+
+struct Hamming : Base {
   template<typename S, typename T>
   struct ANNOY_NODE_ATTRIBUTE Node {
     S n_descendants;
@@ -408,7 +482,8 @@ struct Hamming {
   }
 };
 
-struct Minkowski {
+
+struct Minkowski : Base {
   template<typename S, typename T>
   struct ANNOY_NODE_ATTRIBUTE Node {
     S n_descendants;
@@ -444,7 +519,7 @@ struct Minkowski {
 };
 
 
-struct Euclidean : Minkowski{
+struct Euclidean : Minkowski {
   template<typename S, typename T>
   static inline T distance(const Node<S, T>* x, const Node<S, T>* y, int f) {
     T pp = x->norm ? x->norm : dot(x->v, x->v, f); // For backwards compatibility reasons, we need to fall back and compute the norm here
@@ -478,9 +553,10 @@ struct Euclidean : Minkowski{
   static const char* name() {
     return "euclidean";
   }
+
 };
 
-struct Manhattan : Minkowski{
+struct Manhattan : Minkowski {
   template<typename S, typename T>
   static inline T distance(const Node<S, T>* x, const Node<S, T>* y, int f) {
     return manhattan_distance(x->v, y->v, f);
@@ -559,10 +635,10 @@ protected:
   int _fd;
 public:
 
-  AnnoyIndex(int f) : _f(f), _random() {
-    _s = offsetof(Node, v) + f * sizeof(T); // Size of each node
+  AnnoyIndex(int f) : _f(f + D::internal_dimensions()), _random() {
+    _s = offsetof(Node, v) + _f * sizeof(T); // Size of each node
     _verbose = false;
-    _K = (_s - offsetof(Node, children)) / sizeof(S); // Max number of descendants to fit into node
+    _K = (S) (((size_t) (_s - offsetof(Node, children))) / sizeof(S)); // Max number of descendants to fit into node
     reinitialize(); // Reset everything
   }
   ~AnnoyIndex() {
@@ -570,7 +646,7 @@ public:
   }
 
   int get_f() const {
-    return _f;
+    return _f - D::internal_dimensions();
   }
 
   void add_item(S item, const T* w) {
@@ -586,7 +662,8 @@ public:
     n->children[1] = 0;
     n->n_descendants = 1;
 
-    for (int z = 0; z < _f; z++)
+    memset(n->v, 0, _f * sizeof(T));
+    for (int z = 0; z < (_f - D::internal_dimensions()); z++)
       n->v[z] = w[z];
     D::init_node(n, _f);
 
@@ -600,6 +677,9 @@ public:
       showUpdate("You can't build a loaded index\n");
       return;
     }
+
+    D::template preprocess<T, S, Node>(_nodes, _s, _n_items, _f);
+
     _n_nodes = _n_items;
     while (1) {
       if (q == -1 && _n_nodes >= _n_items * 2)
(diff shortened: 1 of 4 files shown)


erikbern:
This looks nice! One major issue is that the dot product is not a distance function – I'm concerned the splits will be really bad for instance. Let me follow up with some more thoughts though.

erikbern:
@psobot take a look at https://github.com/spotify/annoy/issues/44 – it contains some older notes that may be useful. 

Maybe also https://www.benfrederickson.com/approximate-nearest-neighbours-for-recommender-systems/ although it looks like it refers to the same xbox paper

erikbern:
Very cool!

I don't think you need the `internal_dimensions` thing actually. If you look at the different distance metrics, note that they redefine the `Node` struct. For instance the `Euclidean` distance metric uses that to store an extra element that denotes the offset of the plane for each split. So you could just add an extra element to the `Node` definition for `DotProduct` and use that.

Separately I suspect the preprocessing can be avoided and that you can move that logic into the `create_split` but that's more speculative and I have to think a bit more about it.

erikbern:
> That's one option, although it saves a lot of custom code to just tack on an extra dimension and defer to Angular for create_split and other functions. (Adding an extra element to the Node definition would require us to override those functions to consider the extra element whenever we do math on an entire vector.) If there's a way to do that without adding too much extra complexity though, I'd be interested.

I see that point but I think that's possible to do by just overriding `D::margin` and `D::distance` or something similar right? The `internal_dimensions` is a bit hacky to me – very unlikely to generalize to any other distance metric. I'd rather keep it contained in the `DotProduct` class rather than making it a high level concept

src/annoylib.h
```diff
@@ -934,12 +991,8 @@ template<typename S, typename T, typename Distance, typename Random>
   }
 
   void _get_all_nns(const T* v, size_t n, size_t search_k, vector<S>* result, vector<T>* distances) {
-    Node* v_node = (Node *)malloc(_s); // TODO: avoid
-
-    // Copy the input vector save for the last internal_dimensions, leaving those as zeros.
-    memcpy(v_node->v, v, sizeof(T) * (_f - D::internal_dimensions()));
-    memset(&v_node->v[_f - D::internal_dimensions()], 0, sizeof(T) * D::internal_dimensions());
-
+    Node* v_node = (Node *)calloc(_s, 1); // TODO: avoid
```
erikbern: probably shouldn't mix calloc and malloc :)

erikbern:
nice, starting to look good

src/annoylib.h
```diff
@@ -833,14 +991,15 @@ template<typename S, typename T, typename Distance, typename Random>
   }
 
   void _get_all_nns(const T* v, size_t n, size_t search_k, vector<S>* result, vector<T>* distances) {
-    Node* v_node = (Node *)malloc(_s); // TODO: avoid
-    memcpy(v_node->v, v, sizeof(T)*_f);
+    Node* v_node = (Node *)calloc(_s, 1); // TODO: avoid
```
erikbern: please use malloc here to make it consistent

src/annoylib.h
```diff
@@ -833,14 +991,15 @@ template<typename S, typename T, typename Distance, typename Random>
...
+    memcpy(v_node->v, v, sizeof(T) * _f);
     D::init_node(v_node, _f);
 
     std::priority_queue<pair<T, S> > q;
 
-    if (search_k == (size_t)-1)
-      search_k = n * _roots.size(); // slightly arbitrary default value
+    if (search_k == (size_t)-1) {
+      search_k = D::default_search_k(n, _roots.size());
```
erikbern: why would different distance functions have different implementations? I would rather make it consistent

src/annoylib.h
```diff
@@ -859,7 +1018,7 @@ template<typename S, typename T, typename Distance, typename Random>
         const S* dst = nd->children;
         nns.insert(nns.end(), dst, &dst[nd->n_descendants]);
       } else {
-        T margin = D::margin(nd, v, _f);
+        T margin = D::margin(nd, v_node->v, _f);
```
erikbern: not that it matters but why did you change this?

src/annoymodule.cc
```diff
@@ -263,8 +265,16 @@ py_an_get_nns_by_item(py_annoy *self, PyObject *args, PyObject *kwargs) {
 
 bool
 convert_list_to_vector(PyObject* v, int f, vector<float>* w) {
+  if (PyObject_Size(v) == -1) {
+    char buf[256];
+    snprintf(buf, 256, "Expected an iterable, got an object of type \"%s\"", v->ob_type->tp_name);
+    PyErr_SetString(PyExc_ValueError, buf);
```
erikbern: nice!

src/annoymodule.cc
```diff
@@ -263,8 +265,16 @@ py_an_get_nns_by_item(py_annoy *self, PyObject *args, PyObject *kwargs) {
...
+    return false;
+  }
   if (PyObject_Size(v) != f) {
-    PyErr_SetString(PyExc_IndexError, "Vector has wrong length");
+    char buf[128];
+    snprintf(buf, 128, "Vector has wrong length (expected %d, got %ld)", f, PyObject_Size(v));
```
erikbern: also nice

test/dot_index_test.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def dot_metric(a, b):
+    return -numpy.dot(a, b)
+
+
+def similarity(a, b):
+    # Could replace this with kendall-tau if we're comfortable
+    # bringing in scipy as a test dependency.
+    return float(len(set(a) & set(b))) / float(len(set(a) | set(b)))
```
erikbern: this looks like jaccard similarity to me? anyway why is that needed? in all the other tests we use recall

src/annoylib.h
```diff
@@ -324,7 +347,129 @@ struct Angular {
...
+  }
+
+  static inline size_t default_search_k(const size_t n, const size_t num_roots) {
+    return Angular::default_search_k(n, num_roots) * 10;
+  }
+
+  template<typename S, typename T>
+  static inline T distance(const Node<S, T>* x, const Node<S, T>* y, int f) {
+    return -dot(x->v, y->v, f);
+  }
```
erikbern: shouldn't this just be the euclidean distance between the vectors including the extra element?

src/annoylib.h
```diff
@@ -324,7 +347,129 @@ struct Angular {
...
+    T dot = margin(n, y, f);
+    if (dot != 0)
+      return (dot > 0);
+    else
+      return random.flip();
+  }
+
+  template<typename T>
+  static inline T normalized_distance(T distance) {
+    return distance;
```
erikbern: doesn't entirely make sense to me that this returns the negative dot product. maybe should just return the positive dot product? either way it's not a "distance" in the correct meaning of the word

https://en.wikipedia.org/wiki/Metric_(mathematics)#Definition

src/annoylib.h
```diff
@@ -586,7 +733,8 @@ template<typename S, typename T, typename Distance, typename Random>
     n->children[1] = 0;
     n->n_descendants = 1;
 
-    for (int z = 0; z < _f; z++)
+    memset(n->v, 0, _f * sizeof(T));
```
erikbern: don't think this memset is needed
i guess the for loop could just be replaced by a memcpy though

src/annoylib.h
```diff
@@ -663,7 +815,7 @@ template<typename S, typename T, typename Distance, typename Random>
       // we have mmapped data
       close(_fd);
       off_t size = _n_nodes * _s;
-      munmap(_nodes, size);
+      munmap(_nodes, (size_t) size);
```
erikbern: why did you change this?

if anything should probably change `off_t size` to `size_t size` but i don't remember the difference between `off_t` and `size_t`

erikbern: might just be some win specific thing

erikbern:
This is approved except for my comments

src/annoylib.h
psobot: Fixed!

The reason for changing this to `calloc` was to zero-out the entire Node struct first, but you're right - it's better to be consistent and explicit. I've added a `zero_value` method to `Base` to allow us to set sane defaults for any metrics that require them - in this case, after malloc'ing a new node, we need to set the `dot_factor` of the node to zero.

src/annoylib.h
psobot: 👍 Removed. This was to make the `test_precision` tests easier, as `dot` is a bit less precise than other metrics.

src/annoylib.h
psobot: Ah - this was a holdover from the previous iteration where we used `extra_dimensions`. This is no longer needed. 👍

test/dot_index_test.py
psobot: Switched this to use recall like the other tests.

src/annoylib.h
psobot: I'm not entirely sure, tbh. Both perform fairly well, but if we return `dot` as the actual distance measure, then callers will get back distances that are _actually_ dot products, which might be what callers expect.

I tried switching this to plain cosine distance, and results are pretty similar regardless of which distance metric we use:

```
"dot" with cosine as `distance` (incl extra element): 
Recall at 10 is 70.21%
Recall at 100 is 96.50%
Recall at 1000 is 100.00%

"dot" with dot as `distance`: 
Recall at 10 is 69.55%
Recall at 100 is 96.46%
Recall at 1000 is 100.00%
```

(The reason changing this one method doesn't break _everything_ is that `DotProduct::margin` still takes the `dot_factor` into account.)

src/annoylib.h
psobot: Good point - if we leave this as returning the dot product distance directly, we can flip the sign to return the positive dot product.

src/annoylib.h
psobot: Whoops, this was another holdover from using `extra_dimensions`. I've changed this to a straight `memcpy`.

src/annoylib.h
psobot: Changed this to try and eliminate a bunch of `clang` warnings, but I can undo this.

```src/annoylib.h:846:12: warning: implicit conversion loses integer precision: 'off_t' (aka 'long long') to 'size_t' (aka 'unsigned long') [-Wshorten-64-to-32]```

File: src/annoylib.h (hunk 8 above)
Comment (erikbern) [reply src/annoylib.h]: ok i'm very confused now, let me take a deeper look later today. i think we use the terminology "distance" for multiple separate things.

Comment (yonromai) [reply src/annoylib.h]: (apologies for randomly popping up here and writing an essay)

@erikbern, I don't understand the rational behind using the euclidean distance here?

---
To try to clarify this from an API standpoint:

* The distance function is exposed to the users as follows:
```py
t = AnnoyIndex(2, 'dot')
t.add_item(0, [0, 1])
t.add_item(1, [1, 1])
d = t.get_distance(0, 1)
```
where `get_distance` is defined as follows:
```cpp
T get_distance(S i, S j) {
    return D::normalized_distance(D::distance(_get(i), _get(j), _f));
  }
```

*  The doc does not specify anything special about the distance: "Returns the distance between items `i` and `j`."

* Therefore, I believe the user would expect that `t.get_distance(v_1, v_2)` would use the distance specified in `AnnoyIndex(f, {DIST})`.
So in this case it would be `dot(vectors[v_1], vectors[v_2])`. 

---

One issue is that `dot(x, y)` is more a similarity than a distance. I assume `-dot(...)` is more convenient than `C - dot(...)` since the dot is not normalized.

I guess in this case we could return either `-dot(original_vec_i, original_vec_j)` or `1 - cos(augmented_vec_i, augmented_vec_j)`.
(where `original_vec_i` is of dimension `N` and `augmented_vec_i` is of dimension `N+1`).

Given the API example above, returning the `dot` value seems to make more sense to me than returning the `cos` value.

---

=> @erikbern, @psobot: WDYT?

erikbern: @yonromai the reason i originally brought up euclidean distance was that i thought the distance method was used internally somewhere for the splits.

if it's only for external consumption then i don't have any strong feelings, i guess positive dot product is fine

erikbern:
sorry forgot about this one. will go through shortly – i promise!

erikbern:
haha i forget that this repo is under spotify and you can merge whatever you want :)

i'll try to find some time tonight!

src/annoylib.h
```diff
@@ -225,21 +223,47 @@ inline void two_means(const vector<Node*>& nodes, int f, Random& random, bool co
...
+  static inline void normalize(Node* node, int f) {
+    T norm = get_norm(node->v, f);
+    if (norm > 0) {
+      for (int z = 0; z < f; z++)
+        node->v[z] /= norm;
+    }
+  }
+};
+
+struct Angular : Base {
```
erikbern: this is implementation inheritance, which i'm not a super big fan of. but not blocking

src/annoylib.h
```diff
@@ -324,7 +348,128 @@ struct Angular {
...
+    if (norm > 0) {
+      for (int z = 0; z < f; z++)
+        node->v[z] /= norm;
+      node->dot_factor /= norm;
+    }
+  }
+
+  template<typename S, typename T>
+  static inline T margin(const Node<S, T>* n, const T* y, int f) {
+    return dot(n->v, y, f);
```
erikbern: shouldn't this include the `dot_factor`?

src/annoylib.h
```diff
@@ -324,7 +348,128 @@ struct Angular {
...
+  }
+
+  template<typename S, typename T, typename Random>
```
erikbern: isn't this a reimplementation of the superclass method?

erikbern:
This looks good. I'm curious what would happen if you update the `margin` method as per my suggestion – would be interesting to see if it has an impact on recall!

Side note but Annoy could really use a redesign from scratch at some point...

erikbern:
Also this is just tangentially related but what's the purpose of this PR? I did a fair amount of benchmarking and even though dot product logically makes sense for music recommendation, cosine always performed vastly better. It was always a bit of a mystery to me

src/annoylib.h
psobot: It is, but I'm not sure we can omit it, as `Angular::Node` and `DotProduct::Node` are incompatible types so calling the superclass method directly won't work.

src/annoylib.h
psobot: Good catch - yes, it should! This is fixed.

erikbern:
@psobot so should i merge this?

erikbern:
> Interestingly enough, recall went down by about 2% (at 10 items) in my tests after including dot_factor in the margin calculation. I'm not sure if that means we should leave it out entirely, but if we're trying to match the original Xbox paper, we should include dot_factor for correctness.

that's kind of random btw, hopefully within the level of noise

erikbern:
💥

PR #313, opened 2018-10-06
Inlined all non-static mman functions.

### LINKED ISSUE #312: Multiple definition compilation errors from mman.h on MinGW32
Should the `mmap`, `munmap`, etc. functions in `mman.h` be `inline`d to avoid the ODR? I've been including the `annoylib.h` headers in several `*.cpp` files that are to be used to create a single shared library (in an R package via _RcppAnnoy_, see https://github.com/LTLA/BiocNeighbors) - this works fine, except on Windows where linking [fails](http://bioconductor.org/checkResults/devel/bioc-LATEST/BiocNeighbors/tokay1-install.html) due to multiple definitions of the `mm*` functions. This seems like a standard case for requiring `inline`ing in a header-only library, though I don't have a Windows machine available to confirm it's a fix.

LTLA:
Fixes #312.

### DIFF
--- a/src/mman.h
+++ b/src/mman.h
@@ -85,7 +85,7 @@ static DWORD __map_mmap_prot_file(const int prot)
     return desiredAccess;
 }
 
-void* mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off)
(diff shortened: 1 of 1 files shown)


erikbern:
thanks!

Issue #314, opened 2018-10-07
Catastrophic cancellation with Euclidean distance calculation


LTLA:
The recent updates to the `Euclidean::distance` method at:

https://github.com/spotify/annoy/blob/8729dccb7b4b4a9d3283531dbe5705eb98d52296/src/annoylib.h#L449-L454

... have introduced undesirable behaviour with catastrophic cancellation when `pp`, `pq` and `qq` are all large relative to the final distance between `x` and `y`, _and_ single-precision floats are in use.

To illustrate, we can compile and run this program with the latest `annoylib.h` (as of c1342f2574da88a3433c64fc369e0c247adf16db):

```cpp
#include <cstdio>
#include "annoylib.h"
#include "kissrandom.h"

typedef float ANNOYTYPE;

typedef AnnoyIndex<int, ANNOYTYPE, Euclidean, Kiss64Random> MyAnnoyIndex;

int main() {
    MyAnnoyIndex obj(1);
    
    ANNOYTYPE first=0.7125930, second=0.7123166;
    obj.add_item(0, &first);
    obj.add_item(1, &second);

    obj.build(10);
    
    std::vector<int> stuff;
    std::vector<ANNOYTYPE> outdist;
    obj.get_nns_by_item(0, 2, -1, &stuff, &outdist);

    printf("%i %.10f\n", stuff[0], outdist[0]);
    printf("%i %.10f\n", stuff[1], outdist[1]);
    return 0;
}
```
... which prints out:
```
0 0.0000000000
1 0.0000000000
```
... which is not correct; the distance between items 0 and 1 is equal to 0.0002764. However, if we switch `Euclidean::distance` back to a more direct sum of squares calculation:

```cpp
T d = 0.0;
const T* xp=x->v, *yp=y->v;
for (int i = 0; i < f; i++, xp++, yp++) {
    T tmp=*xp - *yp;
    d += tmp*tmp;
}
return d;
```

... we get a result closer to the truth:

```
0 0.0000000000
1 0.0002764463
```

I know there's a balance between speed and accuracy, but this level of imprecision seems a step too far. I wouldn't consider the above example pathological with respect to numerical precision, yet the program is still failing. Perhaps we should revert back to the old method for computing Euclidean distances. (I suspect at least some of the other distance metrics have similar behaviour but I haven't looked at them in much detail.)

erikbern:
this sounds too large to be attributable to rounding errors. i think there's some arithmetic issue going on.

[...]

PR #316, opened 2018-10-08
More precise Euclidean distance calculations.

LTLA:
Alternative solution to #314 for greater precision and AVX support.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -129,9 +129,11 @@ template<typename T>
 inline T euclidean_distance(const T* x, const T* y, int f) {
   // Don't use dot-product: avoid catastrophic cancellation in #314.
   T d = 0.0;
-  for (int i = 0; i < f; i++) {
-    const T tmp=x[i] - y[i];
+  for (int i = 0; i < f; ++i) {
+    const T tmp=*x - *y;
     d += tmp * tmp;
+    ++x;
+    ++y;
   }
   return d;
 }
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -125,6 +125,17 @@ inline T manhattan_distance(const T* x, const T* y, int f) {
   return d;
 }
 
+template<typename T>
+inline T euclidean_distance(const T* x, const T* y, int f) {
+  // Don't use dot-product: avoid catastrophic cancellation in #314.
+  T d = 0.0;
+  for (int i = 0; i < f; i++) {
+    const T tmp=x[i] - y[i];
+    d += tmp * tmp;
+  }
+  return d;
+}
+
 #ifdef USE_AVX
 // Horizontal single sum of 256bit vector.
 inline float hsum256_ps_avx(__m256 v) {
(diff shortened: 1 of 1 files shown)


erikbern:
interesting. can you check if #315 is enough for what you need though? this PR roughly doubles the amount of arithmetic operations so unless it's truly needed i'd prefer to avoid it

LTLA:
Hmm. I don't want to whale on this, but from my perspective, #315 doesn't seem to be sufficient. 

To demonstrate, here's a little elaboration of the program used in #314:

```cpp
#include <cstdio>
#include "annoylib.h"
#include "kissrandom.h"

typedef float ANNOYTYPE;

typedef AnnoyIndex<int, ANNOYTYPE, Euclidean, Kiss64Random> MyAnnoyIndex;

int main() {
    MyAnnoyIndex obj(1);

    ANNOYTYPE first=1.0001, second=1.0002, third=1.00015;
    obj.add_item(0, &first);
    obj.add_item(1, &second);
    obj.add_item(2, &third);

    obj.build(10);

    std::vector<int> stuff;
    std::vector<ANNOYTYPE> outdist;
    obj.get_nns_by_item(2, 2, -1, &stuff, &outdist);

    printf("%i %.10f\n", stuff[0], outdist[0]);
    printf("%i %.10f\n", stuff[1], outdist[1]);
    return 0;
}
```

On my computer, this gives me:

```
0 0.0000000000
1 0.0000000000
```

.... so not only is the distance wrong, item 3 is not even in its own nearest neighbor set! This seems particularly egregious when the differences between input items are well within `float` precision. I could also imagine a lot of breaking code that assumed the first element of `stuff` would be the item itself.

To be clear; the precision of the output distances themselves is not the problem, as I can just recompute them. The question is how much this loss of precision reduces the accuracy of the NN search, in a manner that is not easily adjustable by increasing `ntrees` or `search_k`. Certainly, I could switch to `double` in the example above (rather unhappily, because of the cost of time and memory), but Python and R users won't be able to do so at all, and I guess that's probably a substantial proportion of Annoy's user base.

Finally - unless I'm reading my code incorrectly - the proposed modification here should only add one subtraction operation plus an extra intermediate variable. Some testing suggests that it's about 50% slower. I'll admit that's not great, but if speed is a priority and accuracy is not, then a more natural way to make that sacrifice would be to reduce `search_k` directly (or indirectly, via a lower `ntrees`).

LTLA:
Cool. I should mention that, in the context of the rest of the Annoy operations, the extra time required for these distance calculations seems negligible. I've tried with a 50000-by-20 random normal matrix, looking for the 20 nearest neighbors with 50 trees, and the performance hit from the switch is 1%.

LTLA:
Nudging this PR. Just to clarify; the 50% slowdown refers to the head-to-head comparison between `dot()` and `euclidean_distance()`. This effect is considerably diluted when considering the Annoy algorithm in its entirety. In my applications, I see a 1-5% slowdown for the search, with more severe performance drops at higher `search_k` (as one might expect, given the context in which `distance()` is called). Nonetheless, this seems like a small price to pay for obtaining a correct final ranking among the collected `search_k` set when the differences between observations are somewhat small.

erikbern:
will take a look shortly. busy life so i don't always have time to jump on pull requests quickly!

erikbern:
Yeah I'm starting to lean towards just reverting my earlier optimization and going back to the (slightly slower) old method. Might as well revert for now so that things are working.

Have been very busy, hopefully if I have half an hour this weekend I'll take a look.

erikbern:
merging this for now. i'm going to manually revert #272 as well

erikbern:
@eddelbuettel take a look at https://github.com/spotify/annoy/pull/325

eddelbuettel:
Perfect, thanks @erikbern. I had approximated this locally.

PR #318, opened 2018-10-13
Add optional parameter to disable prefault

rbares:
Currently Annoy uses the MAP_POPULATE flag when available in calling mmap during ``load`` and ``save``. This attempts to pre-emptively load the entire contents of the index into memory by prefaulting the pages, rather than reading pages as required. This initial memory usage and disk IO can have a negative impact on performance in the following cases:
- available memory is smaller than index size
- few searches are to be performed on an index before it is unloaded
- large areas of the index are unlikely to be relevant to search queries

To mitigate this the ``save`` and ``load`` methods are updated to include an optional prefault argument which can be used to disable MAP_POPULATE. This optional argument defaults to ``true``, providing the existing behaviour. When set to ``false`` MAP_POPUALATE is not used even if available. This provides a trade-off between fast searches with higher memory usage, loading time, and initial disk IO, or slower searches with lower memory usage, loading time, and associated disk IO.

This pull request also includes relevant updates to the Go and Lua bindings, in addition to associated unit tests. The new parameters and their usage are briefly documented in the README.


erikbern:
nice 👍 

i'm tempted to change the default to False actualy

erikbern:
thanks @rbares

rbares:
My pleasure, thanks for the review

Issue #293, opened 2018-04-22
Cannot pip install on Windows 7


okkhoy:
- I downloaded the VC++ 9 compiler + build tools necessary to install Python 2.7 packages. 
- Tried `pip install annoy`

The build fails with the following: `error C3861: '__popcnt64': identifier not found`

```
c:<>\local\temp\pip-build-jmpbob\annoy\src\annoylib.h(350) : error C3861: '__popcnt64': identifier not found
          c:<>\local\temp\pip-build-jmpbob\annoy\src\annoylib.h(714) : see reference to function template instantiation 'T Hamming::distance<S,T>(const Hamming::Node<S,T> *,const Hamming::Node<S,T> *,int)' being compiled
          with
          [
              T=uint64_t,
              S=int32_t
          ]
          c:<>\local\temp\pip-build-jmpbob\annoy\src\annoylib.h(713) : while compiling class template member function 'uint64_t AnnoyIndex<S,T,Distance,Random>::get_distance(S,S)'
          with
          [
              S=int32_t,
              T=uint64_t,
              Distance=Hamming,
              Random=Kiss64Random
          ]
```

there are a whole lot of warnings on improper type conversion in between

```
 error: command 'C:<>\\Local\\Programs\\Common\\Microsoft\\Visual C++ for Python\\9.0\\VC\\Bin\\amd64\\cl.exe' failed with exit status 2

  ----------------------------------------
  Failed building wheel for annoy
```

- Python version:
```
c:\> python --version
Python 2.7.14 :: Anaconda, Inc.
```

eric-kimbrel:
same for windows 10

simon19891101:
Anyone knows how to solve it? Cheers.

tjrileywisc:
This should work for python 3.x, on any OS (I've used it on Win 7 and Win 10 without issues). 

I was never able to get this to build on Python 2.7 unfortunately when I worked on adding Windows support.

rbares:
After testing with a popcnt64 implementation from Sean Anderson's "Bit Twiddling Hacks", handling some further missing identifiers and adding a couple of static casts I finally got this building. Unfortunately my build currently crashes during nosetests but will let you know if I get any further.

erikbern:
@rbares do you have some alternative code for `__popcount64`? i guess we can provide an alternative implementation for visual studio using a preprocessor directive

rbares:
Yes indeed that is the approach in my current implementation, although I have also forced its use in linux for comparison.

[...]

PR #319, opened 2018-10-18
[WIP] Fix python 2.7 windows build

rbares:
Issue #293 reproduced using VC++ 9 to build for Python 2.7 on Windows 10.

- Introduced alternative popcount implementation
- Defined isnan
- Explicitly using std::pow with relevant static_casts

Crash in angular_index_test test_no_items. Test failures in hamming_index_test.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -48,6 +48,7 @@ typedef signed __int32    int32_t;
 #include <algorithm>
 #include <queue>
 #include <limits>
+#include <bitset>
 
 #ifdef _MSC_VER
 // Needed for Visual Studio to disable runtime checks for mempcy
@@ -65,6 +66,9 @@ typedef signed __int32    int32_t;
 
 #ifndef _MSC_VER
 #define popcount __builtin_popcountll
+#elif _MSC_VER == 1500
+#define isnan(x) (sizeof(x) == sizeof(float) ? _isnanf(x) : _isnan(x))
+#define popcount std_popcount
 #else
 #define popcount __popcnt64
 #endif
@@ -406,7 +410,7 @@ struct DotProduct : Angular {
 
   template<typename T, typename Node>
   static inline void normalize(Node* node, int f) {
-    T norm = sqrt(dot(node->v, node->v, f) + pow(node->dot_factor, 2));
+    T norm = sqrt(dot(node->v, node->v, f) + std::pow(node->dot_factor, 2));
     if (norm > 0) {
       for (int z = 0; z < f; z++)
         node->v[z] /= norm;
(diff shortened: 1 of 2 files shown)


test/angular_index_test.py
```diff
@@ -208,6 +208,7 @@ def test_only_one_item(self):
         self.assertEquals(idx.get_n_items(), 1)
         self.assertEquals(idx.get_nns_by_vector(vector=numpy.random.randn(100), n=50, include_distances=False), [0])
 
+    @SkipTest # TODO: fix crash
```
erikbern: is there an easy way for you to figure out why it's crashing? it doesn't crash on linux and in appveyor

src/annoylib.h
```diff
@@ -488,6 +492,10 @@ struct Hamming : Base {
   static inline T pq_initial_value() {
     return numeric_limits<T>::max();
   }
+  template<typename T>
+  static inline int std_popcount(T v) {
```
erikbern:
```suggestion
  static inline int std_popcount(T v) {
    // Note: this is only used with Visual Studio
```

erikbern: might want to add a comment like ^

erikbern:
Thanks – appreciate the contribution

test/angular_index_test.py
rbares: Will be trying the VS mixed-mode debugger this weekend. From additional logging the crash occurs in https://github.com/spotify/annoy/blob/a7b351222cc0ab25ac1ecd6282c1d26836d3a003/src/annoylib.h#L932 when ``indices.size() == 0``. I can't imagine why as that sounds equivalent to a nop, but conditionally executing the memcpy fixes the test.

src/annoylib.h
rbares: Looks good. I'm unclear on whether this will use intrinsics or not so this make have a performance impact.

test/angular_index_test.py
erikbern: maybe just doing a `if(!indices.empty())` gate would resolve it on all platforms

weird though

erikbern: oh i bet `indices[0]` will cause a memory access violation if indices has zero size. that's funny.

easy fix though. i can do it myself.

rbares: Yes that's what I used. The &index[0] should be fine as we only ever get the address, so I suspect a bug in this old memcpy implementation may be causing an unnecessary attempt to read

erikbern:
something looks weird with the diff:

image

any idea what's going on?

anyway thanks for this change!!

rbares:
> something looks weird with the diff:
> 
> image
> 
> any idea what's going on?
> 
> anyway thanks for this change!!

Oh dear. I allowed github to commit the comment you proposed using the web interface ( https://github.com/spotify/annoy/pull/319/commits/7650312212c5c193dc485dc346fee12a48c9f829 ). It looks like that has changed line endings throughout the file. My apologies. Will sort that out while debugging the remaining tests.

erikbern:
haha that's funny. i thought maybe it was something with CR+LF since this was a windows related fix. wasn't too far off.

rbares:
Ready for review

PR #297, opened 2018-05-06
reduce realloc overhead, MacOS compile fix

ViperCraft:

PR #323, opened 2018-10-27
remove need for malloc in a bunch of places

erikbern:
idea stolen from #297 

use stack allocations instead


LTLA:
Is `char node_buffer[s*2];` legal C++? It doesn't seem that `s` is a compile-time constant, and variable length arrays don't seem to be standard (though they are supported by GCC), see [here](https://stackoverflow.com/questions/15013077/arrayn-vs-array10-initializing-array-with-variable-vs-real-number) and [here](https://stackoverflow.com/questions/1887097/why-arent-variable-length-arrays-part-of-the-c-standard).

barabanus:
Replace `malloc` with `alloca`

erikbern:
thanks @barabanus – appreciate it!

PR #329, opened 2018-10-31
test for msc_ver and mingw

eddelbuettel:
This reflects the change I had to make for the most recent RcppAnnoy upload:

- for R the Windows compiler is always MinGW's `gcc`;
- we need to test for it to in order to get `mman.h` included;
- we also already have `NOMINMAX` defined so this gets an extra test.


erikbern:
thanks @eddelbuettel !

Issue #335, opened 2018-11-29
Bus error in get_nns_by_vector


mxgr7:
We are seeing bus errors in production from time to time when calling `get_nns_by_vector`. The process crashes with SIGBUS. We are stuck because we can't answer the following questions:

1. We are not able to reproduce it in a development environment because we have no idea what could be causing it. Are there any obvious scenarios that would explain the SIGBUS?
2. The core dumps from production don't help us either because we can't get gdb to use the correct symbols and show us function names etc. Is there a way to get that to work without running a debug version of Python in production?

Some more context: 
* We are running two processes, both loading the same index files from disk into memory. 
* After the initial loading at startup, one of them becomes the "query" process which handles incoming requests by searching the loaded indices. 
* The other becomes the "fetch" process which monitors the file system for changes to index files and reloads any changed index. 
* Once the fetch process finishes reloading an index, the roles reverse and it becomes the "query" process while the other becomes the "fetch" process.
* These two processes are started from a main process which runs Flask and which is in turn started by uwsgi inside a docker container. uwsgi is configured to use a single process and a single thread.

At this point we are running out of ideas and any pointers on what to look at would be greatly appreciated.

erikbern:
Interesting. No idea what's going on.

> The other becomes the "fetch" process which monitors the file system for changes to index files and reloads any changed index.

This seems fishy. I'm not sure what happens if you mmap a file and then the file changes. I wouldn't be surprised if it leads to complications. Might make more sense to create a new file, but never modify existing files.

> Once the fetch process finishes reloading an index

How do you reload the index? Again, a bit nervous about reloading. You should probably just create a new index and load a new file from scratch.

mxgr7:
Thank you!

> I'm not sure what happens if you mmap a file and then the file changes.

We had not considered that, but you raise a valid point! If this actually causes problems it could explain the observed failures. Will investigate further.

> How do you reload the index?

We initialize a new `AnnoyIndex` object and call `.load()` to read the file from disk.

erikbern:
> Will investigate further.

nice, let me know what you find!

> We initialize a new `AnnoyIndex` object and call `.load()` to read the file from disk.

Ok, this should be fine.

erikbern:
https://stackoverflow.com/questions/12383900/does-mmap-really-copy-data-to-the-memory

You might definitely get issues if you modify existing files. However the way the Linux kernel works (IIRC) is if you delete the file and create a new one with the same filename, that's considered a different file.

[...]

PR #339, opened 2018-12-05
Delete existing file before saving

erikbern:
One liner fix to #335


erikbern:
Don't think this warrants a new release so I'll wait a bit until I upload a new version to PyPI

PR #361, opened 2019-02-20
Recommend `pip install --user` over `sudo pip install` in README

Calinou:
This removes the need for users to enter their password when installing the package and is safer.


erikbern:
Thanks @Calinou

Issue #379, opened 2019-04-08
save() ignores write failures, causing corrupt annoy trees


keithmcneill:
In particular if the disk is full, the annoy tree will be silently truncated.

Working on a fix on a fork.

erikbern:
thanks for the fix!

PR #380, opened 2019-04-08
Fix for #379, save() ignores write failures

keithmcneill:


test/index_test.py
```diff
@@ -172,3 +172,35 @@ def test_get_n_trees(self):
...
+        f = 40
+
+        # Build the initial index
+        t = AnnoyIndex(f)
+        for i in range(1000):
+            v = [random.gauss(0, 1) for z in range(f)]
+            t.add_item(i, v)
+        t.build(10)
+
+        if sys.platform == "linux" or sys.platform == "linux2":
```
psobot: nit: [might want to do `if sys.platform.startswith('linux')` here](https://stackoverflow.com/a/10429736/679081).

src/annoylib.h
```diff
@@ -878,11 +880,25 @@ template<typename S, typename T, typename Distance, typename Random>
       if (f == NULL)
         return false;
 
-      fwrite(_nodes, _s, _n_nodes, f);
-      fclose(f);
+      bool write_failed = false;
```
erikbern: don't think this one is really needed, could just `return false` at each point?

not blocking though

erikbern:
thanks!

test/index_test.py
```diff
@@ -172,3 +172,35 @@ def test_get_n_trees(self):
...
+            # linux
+            try:
+                t.save("/dev/full") 
+                self.fail("didn't get expected exception")
+            except Exception as e:
+                self.assertTrue(str(e).find("No space left on device") > 0)
```
psobot: Python nit: could do `self.assertTrue("No space left on device" in str(e))` or in newer versions, `self.assertIn("No space left on device", str(e))`

test/index_test.py
```diff
@@ -172,3 +172,35 @@ def test_get_n_trees(self):
...
+        elif sys.platform == "darwin":
+            volume = "FULLDISK"
+            device = os.popen('hdiutil attach -nomount ram://64').read()
+            os.popen('diskutil erasevolume MS-DOS %s %s' % (volume, device))
+            os.popen('touch "/Volumes/%s/full"' % volume)
+            try:
+                t.save('/Volumes/%s/annoy.tree' % volume)
+                self.fail("didn't get expected exception")
+            except Exception as e:
+                self.assertTrue(str(e).find("No space left on device") > 0)
```
psobot: Same here.

test/index_test.py
```diff
@@ -172,3 +172,35 @@ def test_get_n_trees(self):
...
+            # linux
+            try:
+                t.save("/dev/full") 
+                self.fail("didn't get expected exception")
+            except Exception as e:
+                self.assertTrue(str(e).find("No space left on device") > 0)
+        elif sys.platform == "darwin":
+            volume = "FULLDISK"
+            device = os.popen('hdiutil attach -nomount ram://64').read()
+            os.popen('diskutil erasevolume MS-DOS %s %s' % (volume, device))
```
psobot: This works, but I'm pretty wary of calling `diskutil erasevolume` from a unit test. Is there a simpler way we can cause an error? (Maybe writing to a special file, like a named pipe or socket instead?)

src/annoylib.h
keithmcneill: I refactored that a few times deciding if I wanted to unload or not in case of error.  That is a vestige of that.

keithmcneill:
@erikbern When do you expect you're going to push out a new version?  This fix is a for a production incident I'd like to get sorted out.  Thanks!

erikbern:
I can do that later tonight, but please remind me if I forget!

erikbern:
The test doesn't seem to be working on OS X:

```
touch: /Volumes/FULLDISK/full: No such file or directory
F
FAIL: test_write_failed (index_test.IndexTest)
  File "/Users/erikbern/annoy/test/index_test.py", line 199, in test_write_failed
    t.save('/Volumes/%s/annoy.tree' % volume)
FileNotFoundError: [Errno 2] No such file or directory

During handling of the above exception, another exception occurred:

  File "/Users/erikbern/annoy/test/index_test.py", line 202, in test_write_failed
    self.assertTrue(str(e).find("No space left on device") > 0)
AssertionError: False is not true

Ran 1 test in 1.370s
```

keithmcneill:
Interesting, I wonder if one of the previous steps failed, and then didn't actually create the volume.

Given that there is the linux test, I think you can just delete the mac test.

erikbern:
ok, will just delete it

keithmcneill:
@erikbern Any chance of getting out a new version this week so I can update our internals here?

Keith

erikbern:
sure can do tonight

erikbern:
done

Issue #398, opened 2019-07-06
Adding items to existing index causes Segmentation Fault


Jotschi:
Example:
```
#!/usr/bin/env python3

from annoy import AnnoyIndex
import random
import os  

indexName = "index.ann"
VECTOR_LENGTH = 32

i = 1
u = AnnoyIndex(VECTOR_LENGTH, metric='hamming')

if os.path.exists(indexName):
  print("Loading index " + indexName)
  u.load(indexName)  
  i=i+1

v = [random.gauss(0, 1) for z in range(VECTOR_LENGTH)]
print("Adding item " + str(i))
u.add_item(i, v)


print("Building tree")
u.build(10)
print("Saving index")
u.save(indexName)

print(u.get_nns_by_item(0, 1000))
```

`rm index.ann ; ./db.py ; echo -e  "\nSecond:" ; ./db.py`

Calling the program the first time works just fine. However when calling `u.add_item` on the loaded index a `Segmentation fault` is observed.

OS: Debian Linux
Python: 3.7.3-2
Annoy: 1.15.2

erikbern:
Annoy doesn’t let you add items to existing indexes. That being said, it probably shouldn’t segfault. I’ll fix

ljeagle:
@Jotschi  we have built a system named [vearch](https://github.com/vearch/vearch), which can support concurrent add when searching. If you have the need, you can have a try. Thanks

micah5:
This issue seems to be resolved, you currently get:
```
Exception: You can't add an item to a loaded index
```

PR #399, opened 2019-07-07
raise exceptions instead of segfaulting or doing nothing

erikbern:
should address #398 and possibly some other issues


erikbern:
in general i'm hoping to throw exceptions rather than segfaulting in every case

there's a few OOB cases missing (`get_item` etc) but this should cover some common user errors

PR #400, opened 2019-07-10
update readme python example

andrewnc:
change `xrange` -> `range` so code example works in both python2 and python3.


erikbern:
awesome, appreciate it @andrewnc !

PR #418, opened 2019-09-21
assign ftruncate result to quieten g++

eddelbuettel:
While updating RcppAnnoy to the current annoy version, I noticed `g++` whining about this under the settings we use -- it is technically an unassigned return value so now we assign it.  To then ignore it :)

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -1121,7 +1121,7 @@ protected:
       void *old = _nodes;
       
       if (_on_disk) {
-        ftruncate(_fd, _s * new_nodes_size);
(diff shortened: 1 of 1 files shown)


erikbern:
thanks @eddelbuettel !

eddelbuettel:
Anytime!  The changes between 0.12 (where I was) and 0.16 (where I am now) look awesome.  How much did the newer vectorised code help (on suitable hardware) ?

erikbern:
It did speed things up quite a lot! I think almost 4x

PR #419, opened 2019-09-22
use return code, and only conditionally define ftruncate

eddelbuettel:
Follow-up to yesterday's #418 while preparing CRAN upload:

1.  Need to actually use return code from `ftruncate`.  Doh.  So added minimal output, conditional on `_verbose` being set as well.

2.  Turns out the Windows version of `g++` does not like the new definition of `ftruncate` as it has its own.  So `#if !defined` wrapped around it.

With that I am good.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -1122,6 +1122,7 @@ protected:
       
       if (_on_disk) {
         int rc = ftruncate(_fd, _s * new_nodes_size);
+        if (_verbose && rc) showUpdate("File truncation error\n");
         _nodes = remap_memory(_nodes, _fd, _s * _nodes_size, _s * new_nodes_size);
       } else {
         _nodes = realloc(_nodes, _s * new_nodes_size);
--- a/src/mman.h
+++ b/src/mman.h
@@ -209,6 +209,7 @@ inline int munlock(const void *addr, size_t len)
     return -1;
 }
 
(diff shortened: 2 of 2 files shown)


erikbern:
These both seem a bit hacky. Probably would make more sense to make `_allocate_size` return a `bool` and throw an error on it. For the second point feels like there should be a more granular macro. 

But good enough to merge for now, can look into it later!

eddelbuettel:
Actually we now use a differentiation for 'yes we are on Windows but there is more than just using MSVC' in one other place: [line 38 in annoylib,h](https://github.com/spotify/annoy/blob/master/src/annoylib.h#L38) and I am actually quite grateful for it.  What we need here may just be what I added:  yes, add `mman.h` as we are on Windows but no, do not redefine `ftruncate` as MinGW brings it own (with a more POSIXy signature using `off_t` for size).   

As for the reworking of `_allocate_size()`: sounds like a good idea.  I just wanted something minimal to proceed.

(Also, turns out reverse dependency checks at CRAN threw a spanner on save/load comparison.  I'll CC you in email in a minute.  Could be anywhere between you upstream, me in the middle or that user downstream.)

erikbern:
your PR is totally good for a merge for now. i just made a note of some things worth fixing later!

thanks!

PR #426, opened 2019-09-28
Handle lseek failures – follow up from #423

### LINKED ISSUE #423: OSError: Index size is not a multiple of vector size
i  use  annoy to build ,which the number of words is 5844240 and the vector size is 200. it raise a error: 【OSError: Index size is not a multiple of vector size】。please help me

erikbern:
Doesn't resolve #423, but will make error cases with lseek more transparent

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -1042,8 +1042,10 @@ public:
       return false;
     } else if (size == 0) {
       showUpdate("Size of file is zero\n");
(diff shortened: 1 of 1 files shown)


erikbern:
need to fix a few wacky cases to make this build

erikbern:
this unveiled one or maybe two other issues that i'll fix separately

Issue #61, opened 2015-05-06
Segfault if you save() without doing build()


a1k0n:
It's a PEBKAC error but it's sort of user unfriendly.

```
Program received signal SIGSEGV, Segmentation fault.
*__GI___libc_free (mem=0x7fff7974d000) at malloc.c:3710
3710    malloc.c: No such file or directory.
        in malloc.c
(gdb) where
#0  *__GI___libc_free (mem=0x7fff7974d000) at malloc.c:3710
#1  0x00007fffefff95bc in AnnoyIndexPython<float, Angular<float> >::save_py(std::string const&) () from /usr/lib/pymodules/python2.6/annoy/annoylib.so
#2  0x00007fffefff44b7 in boost::python::objects::caller_py_function_impl<boost::python::detail::caller<void (AnnoyIndexPython<float, Angular<float> >::*)(std::string const&), boost::python::default_call_policies, boost::mpl::vector3<void, AnnoyIndexPython<float, Angular<float> >&, std::string const&> > >::operator()(_object*, _object*) ()
   from /usr/lib/pymodules/python2.6/annoy/annoylib.so
#3  0x00007fffefdb7bde in boost::python::objects::function::call(_object*, _object*) const () from /usr/lib/libboost_python-py26.so.1.42.0
#4  0x00007fffefdb7e88 in ?? () from /usr/lib/libboost_python-py26.so.1.42.0
#5  0x00007fffefdbf703 in boost::python::detail::exception_handler::operator()(boost::function0<void> const&) const () from /usr/lib/libboost_python-py26.so.1.42.0
#6  0x00007fffefff2c67 in boost::detail::function::function_obj_invoker2<boost::_bi::bind_t<bool, boost::python::detail::translate_exception<ErrnoException, void (*)(ErrnoException const&)>, boost::_bi::list3<boost::arg<1>, boost::arg<2>, boost::_bi::value<void (*)(ErrnoException const&)> > >, bool, boost::python::detail::exception_handler const&, boost::function0<void> const&>::invoke(boost::detail::function::function_buffer&, boost::python::detail::exception_handler const&, boost::function0<void> const&) ()
   from /usr/lib/pymodules/python2.6/annoy/annoylib.so
#7  0x00007fffefdbf4d3 in boost::python::handle_exception_impl(boost::function0<void>) () from /usr/lib/libboost_python-py26.so.1.42.0
[...]

PR #427, opened 2019-09-29
Disallow saving an index that hasn't been built

erikbern:
This was de facto not working, and to support it would have been kind of annoying. I think it was working at some point (see #61) but it must have broken later. Also seems like kind of a pointless feature to have.

Better to just be explicit that it doesn't work.

This will also let me fix the lseek warning messages properly.


erikbern:
looks like appveyor is down. merging since i don't see why this would affect win differently

PR #429, opened 2019-09-30
naive attempt at adding os x

erikbern:
i've tried this in the past and failed (#118) but maybe it's working now?


erikbern:
awesome. noticed travis has beta support for win too now, maybe worth looking into

would be nice to build wheels on these

PR #431, opened 2019-10-18
Allow configuring target compile architecture as install-option

NegatioN:
Hi.

We're in a position of compiling a docker image containing Annoy on a build-server with a different architecture then our target server has. This may result in Annoy compiling with instructions that are unavailable at the target architecture. It would be super helpful if we could set this option with pip instead of precompiling Annoy somewhere else.

I've tested this locally like this: `pip install -v . --install-option="--march=knl"`. Everything seems to compile alright. 
Also, a regular install works as expected and should give the same compiler options as before `pip install -v .`

I get a `ModuleNotFoundError: No module named 'annoy.annoylib'` when trying to import my locally compiled version of both master, and my own build, so I'm pretty sure this is just my own pip-noobing showing it's face and not something breaking.

Having this resolved quickly would make our lives a lot easier :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ from setuptools import setup, Extension
 import codecs
 import os
 import platform
-import sys
+from setuptools.command.install import install as _install
 
 readme_note = """\
 .. note::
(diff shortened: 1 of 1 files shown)


NegatioN:
@erikbern Is this something which is feasible to get merged?

erikbern:
I think I'm cool merging this, but it looks super hacky – is there no cleaner way to accomplish the same thing?

NegatioN:
After struggling a long time without seeing a simple way to configure this, it struck me that we can just use environment variables. I think this is the cleanest approach, which allows full customization without breaking any existing installations.

It now takes in a comma-separated list of compiler and linker args, and if set will override anything specified.

Would you like some documentation for this, or should it just be on a "Need to know"-basis?

erikbern:
I'm OK with this – seems like a bit of an edge case anyway, so i don't think we need any documentation.

Curious what you're using it for!

NegatioN:
Thanks for the quick merge :smile:

We're bulding docker images used for model training on in-house build servers. They seem to be running on a `knl` architecture, giving us a library compiled with instructions not available at our actual model-training servers, which are running on good old `haswell` :)

Frankly I'm surprised more libraries aren't broken as a result of this. Seeing as there are quite a few that compile C.

For now I'll just install via git, but do you have any plans releasing a new minor version in the near future?

erikbern:
i just released a new version of annoy and there's just a few very minor changes since then, so it might be a while until next version, maybe a month? hopefully that's ok

NegatioN:
Sounds okay :) We'll keep a watch out. Thanks again.

Issue #432, opened 2019-10-21
Failing angular distance tests


karlhigley:
When I clone master at v1.16.2 and run the tests, I get two failures with Python 2.7.16:

```
======================================================================
FAIL: test_distance_consistency (angular_index_test.AngularIndexTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/karl/Projects/ml/annoy/test/angular_index_test.py", line 196, in test_distance_consistency
    self.assertAlmostEqual(dist, numpy.dot(u_norm - v_norm, u_norm - v_norm) ** 0.5)
  File "/Users/karl/Projects/ml/annoy/test/common.py", line 20, in assertAlmostEqual
    super(TestCase, self).assertAlmostEqual(x, y, 3)
AssertionError: 0.0005146485636942089 != 0.0 within 3 places

======================================================================
FAIL: test_single_vector (angular_index_test.AngularIndexTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/karl/Projects/ml/annoy/test/angular_index_test.py", line 226, in test_single_vector
    self.assertEquals(a.get_nns_by_vector([1, 0, 0], 3, include_distances=True), ([0], [0.0]))
AssertionError: Tuples differ: ([0], [0.0003452669770922512]) != ([0], [0.0])

First differing element 1:
[0.0003452669770922512]
[0.0]

- ([0], [0.0003452669770922512])
+ ([0], [0.0])

----------------------------------------------------------------------
Ran 99 tests in 175.784s

FAILED (failures=2)
```

Any ideas why these tests would be failing? (With Python 3.7.4, only the second test fails.)

The Python 2 build logs look like this:

```
$ python setup.py nosetests
running nosetests
running egg_info
writing annoy.egg-info/PKG-INFO
writing top-level names to annoy.egg-info/top_level.txt
writing dependency_links to annoy.egg-info/dependency_links.txt
reading manifest file 'annoy.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'annoy.egg-info/SOURCES.txt'
building 'annoy.annoylib' extension
creating build/temp.macosx-10.14-x86_64-2.7
creating build/temp.macosx-10.14-x86_64-2.7/src
[...]

PR #433, opened 2019-10-21
Fix some angular distance tests related to `/` and `sqrt` rounding error

karlhigley:
These tests were failing when run on OSX 10.14 with Python 2.7.16. The
latter test also failed with Python 3.7.4. They pass with these changes.


karlhigley:
Fixes #432.

PR #435, opened 2019-10-22
Replace `malloc` with `alloca` to allocate on the stack

karlhigley:
Resolves #323 by following a suggestion from @barabanus.


karlhigley:
Not sure if `malloc` in the [go module](https://github.com/spotify/annoy/blob/master/src/annoygomodule.i#L40) should also change...

erikbern:
oh awesome, nice idea

PR #436, opened 2019-10-24
Add `numpy` and `h5py` as dependencies in setup.py's `tests_require`

karlhigley:
`numpy` and `h5py` are required to run the tests but those dependencies
aren't currently specified anywhere AFAIK. When running the tests via
`python setup.py nosetests`, this will use any existing installed
packages (e.g. via pip) or download them to the `.eggs` directory if not
already installed. It makes the tests "just work" the first time you try
to run them after cloning the repo.


erikbern:
thanks @karlhigley

PR #442, opened 2019-12-17
Fix #388 issue – Windows failure with large indexes

### LINKED ISSUE #388: Rewrite existing file issue
Hello! I'm using annoy 1.15.2 on python 3.7.3. Here is the issue:
```
>>>from annoy import AnnoyIndex
>>>a=AnnoyIndex(1)
>>>a.add_item(1,[1])
>>>a.add_item(2,[2])
>>>a.build(1)
True
>>>a.save("temp.ann")
True
>>>b=AnnoyIndex(1)
>>>b.add_item(3,[3])
>>>b.add_item(4,[4])
>>>b.build(1)
True
>>>b.save("temp.ann")
  File "<input>", line 1, in <module>
OSError: [Errno 22] Invalid argument
```
Could you tell why I get the error?

erikbern:
So far this is just a unit test. This runs well on my local computer (Mac) but should reproduce the failure in #388. I expect this to fail on Windows.

Removed a debug output left by accident earlier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
     - os: osx
       language: generic
       env:
-        - TOXENV=py36
+        - TOXENV=py37
 
 env:
   matrix:
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -1044,7 +1044,11 @@ public:
       _fd = 0;
       return false;
     }
+#ifndef _MSC_VER
     off_t size = lseek(_fd, 0, SEEK_END);
+#else
+    off_t size = _lseeki64(_fd, 0, SEEK_END);
+#endif
     if (size == -1) {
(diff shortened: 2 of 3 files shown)


erikbern:
Now using _lseeki64 on windows. Let's see if it runs on Appveyor

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/lseek-lseeki64?view=vs-2019

erikbern:
This now seems to handle fseek, but something is still broken. It's possible mmap is fundamentally not working with 2GB+ files in Windows.

erikbern:
looks like the issue was that off_t is defined as 32 bit in windows by default. i'm now overriding it to be 64 bit in a hacky way. think this should work!

erikbern:
Nice, looks like it’s working now!

PR #443, opened 2019-12-18
include strerror in the exception itself

erikbern:
pure ergonomics, just makes the exception messages a bit easier.

`OSError: Unable to open: No such file or directory (2)`

instead of 

`OSError: Unable to open`

etc

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -83,7 +83,7 @@ void set_error_from_errno(char **error, const char* msg) {
 void set_error_from_string(char **error, const char* msg) {
   showUpdate("%s\n", msg);
   if (error) {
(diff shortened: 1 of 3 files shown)


src/annoylib.h
```diff
@@ -985,20 +994,17 @@ template<typename S, typename T, typename Distance, typename Random>
 
       FILE *f = fopen(filename, "wb");
       if (f == NULL) {
-        showUpdate("Unable to open: %s\n", strerror(errno));
-        if (error) *error = strerror(errno);
+	set_error_from_errno(error, "Unable to open");
```
erikbern: something w tabs vs spaces is messed up here, should fix in a separate PR

PR #444, opened 2019-12-19
Inlines ftruncate

thoellt:
When compiling with MSVC 2013 linking fails when annoy is included in multiple classes.
Inlining it fixes the issue.


erikbern:
odd but 👍

erikbern:
thanks! there might be a way to set up appveyor to build with different versions of MSVC but i'll leave it for another time to investigate

PR #457, opened 2020-02-24
Fix 456 (unaligned memory for Hamming)

erikbern:
This might fix it, @jlmelville wdyt?


jlmelville:
Thanks for attempting this so quickly @erikbern. This passes the GCC-UBSAN check on Rhub, but it's hard to be certain of the fix because due to the stochastic nature of the original problem.

erikbern:
cool. i'll try to address the failing travis build in the near future

jayvdb:
note this causes segfaults in `hamming_index_test.py::HammingIndexTest::test_many_vectors`

erikbern:
i completely forgot about this PR, will revisit shortly

PR #460, opened 2020-02-29
use int for search_k rather than size_t

erikbern:
minor fix to deal with a sanitizer issue in https://github.com/eddelbuettel/rcppannoy/issues/56

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -828,8 +828,8 @@ class AnnoyIndexInterface {
   virtual void unload() = 0;
   virtual bool load(const char* filename, bool prefault=false, char** error=NULL) = 0;
   virtual T get_distance(S i, S j) const = 0;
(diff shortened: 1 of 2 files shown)


erikbern:
@eddelbuettel does this look right to you?

eddelbuettel:
Just back in from a day of conferencing.  Looks pretty promising!

eddelbuettel:
We also have another (minor) change (I missed) earlier in that we need an `inline` tag in front of two helper functions because folks may include the header multiple times.  See https://github.com/eddelbuettel/rcppannoy/pull/58 which does it but I think you were already in a thread or two where I suggested what @LTLA then quickly put into a PR.

erikbern:
thanks, i'll inline those things in a separate pr

eddelbuettel:
Looks good too.  Builds with both PRs, still get some g++-9 line noise. But hopefully the heavier SAN stuff is under control.

PR #466, opened 2020-03-17
Add documentation for set_seed

### LINKED ISSUE #465: Exposing _random in AnnoyIndex to python
I'd like to be able to have more control about the randomness when using annoy.  In essence I'd love to be able to pass a seed, or even better, an instance of numpy.random.RandomGenerator, although a seed will most likely suffice.

I think editing [this line](https://github.com/spotify/annoy/blob/20047ca84b44f471c5739edd6ea7f22484e80fe1/src/annoylib.h#L846) and moving it to the public part would be the first step.

What I'm still unclear on is how the C++ object would relate to the corresponding python object (I assume this is nontrivial). Furthermore, can the `Random` object even be initialized with an int?  In general how Python and C++ work together is still a bit of a mystery to me.

jnboehm:
Fixes #465

This function was already available for use in python but not
documented/mentioned before.


erikbern:
THanks!

PR #478, opened 2020-05-03
A more informative error for  #423

thvasilo:
Users might commonly save with a non-default metric, then try to open using the default metric. This gives a more informative message.

@erikbern let me know if you like the language and if line length is a concern.


erikbern:
Thanks. It could also be because vector length changed, let me add a comment about that as well. Thanks!

PR #494, opened 2020-07-24
Annoy/make ftruncate work on windows

vbod:
## Functional
- Before: `ftruncate()` function would crash for big files and `on_disk_build()` also.
- Now: should work 🤞 

## Technical
Change the Windows API calls for more appropriate ones when dealing with large files. Also changed the order of unmap / map / truncate that used to be done while ajusting file size to unmap / truncate / map in order to avoid the ERROR_USER_MAPPED_FILE (1224).

## Tests
- Ran all unit-tests OK
- Used a dataset that used to crash on my Windows-10 (x64) and ensure it does work now.

## Help
I have very restricted access to other OS than Windows right now... It might be good to double-check I did not crash all of the others 😉 

## ToDo
I use `fprintf()` inside `mman.h` which is not redirected to a custom user printer, it might be implemented if need be (though I thought it was useful as a user to have a log of what kind of error occurred in a more readable way than just "I/O error" but did not bother factorizing the logger system to avoid duplicating the macro... ).

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -123,6 +123,12 @@ inline void set_error_from_string(char **error, const char* msg) {
 #endif
 #endif
 
+#if !defined(__MINGW32__)
+#define FTRUNCATE_SIZE(x) static_cast<int64_t>(x)
+#else
+#define FTRUNCATE_SIZE(x) (x)
+#endif
+
 
 using std::vector;
 using std::pair;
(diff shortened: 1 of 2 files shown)


erikbern:
looks good!

erikbern:
is this ready to be merged?

vbod:
Yes ready to merge 👍

erikbern:
Awesome thanks @vbod!

Issue #70, opened 2015-05-29
Parallelize build?


ravimody:
Conceptually, if the trees are independent, shouldn't we be able to build the trees in parallel across multiple cores? Is this something that could be implemented easily? 

I haven't had a chance to dive deeply into the code/algorithm yet so I may be misunderstanding something.

erikbern:
It wouldn't be too hard – you would need a mutex in a couple of places and also watch out for reallocations, but other than that it should be easy

ravimody:
Cool thanks, I'll look into it (although my C++ is a little rusty :))

thomas4g:
Has there been any further work on this?

erikbern:
not afaik :(

ravimody:
No work on it from my end.

thomas4g:
Bummer. I'd love to help out, but I don't think I'm familiar enough with annoy yet (or frankly competent enough with C++) to spearhead anything. If anyone starts working on this and wants a hand, I'm happy to help out.

thomas4g:
Actually, I take it back. It looks like a pretty simple change. If I understand it properly, it's line 496 of `annoylib.h`:

    _roots.push_back(_make_tree(indices));

That can get parallelized, but with a mutex around `_roots` to avoid simultaneous `push_back` calls, right?

erikbern:
roughly, but  it's a bit more complicated.`_make_tree` also modifies the datastructure. You also need a mutex around line 643-644 and line 702-703. Maybe that mutex could just be folded into `_allocate_size` and we could rename the function to something like `_add_item` ... probably makes more sense

then you obviously need to create pthreads and collect them afterwards... honestly I haven't done that in like 10 years, but iirc it's not too hard

tjrileywisc:
@erikbern Actually I was working on a branch to add this just this weekend. It doesn't quite work yet though (I can only get it to run the precision_test.cpp example, only a debug binary works and it crashes sometimes even then); should I submit a PR?

I didn't touch anything towards the bottom of annoylib.h (near those lines that you mentioned), that might be the problem...

erikbern:
sure, feel free to submit a PR, just make sure to put WIP in the subject or something

you definitely need a mutex around those lines, that's probably the issue :)

thomas4g:
[...]

PR #495, opened 2020-07-28
Multithreaded build

novoselrok:
Issue: https://github.com/spotify/annoy/issues/70

This is my work-in-progress branch for supporting multithreaded builds in Annoy. Locally (Ubuntu 18.04) all C++, Python and Lua tests seem to pass (I've re-ran them multiple times). The only thing remaining is fixing the Go build and I would appreciate some help with this :smile: 

I've ran some speed tests on glove-angular-25 index with n_trees=32. Reference time on master (without threads) was 49.9s.
Here are results for multiple threads:
| Num. threads | Time (s)    |
|--------------|-------------|
|            1 | 51.28759694 |
|            2 |   26.675179 |
|            4 | 14.82837749 |
|            8 | 8.897883892 |
|           16 | 7.353658915 |

If you have any questions or concerns feel free to comment.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -57,9 +57,9 @@ typedef signed __int64    int64_t;
 #include <algorithm>
 #include <queue>
 #include <limits>
-#include <mutex>
 #include <thread>
-#include <condition_variable>
+#include <mutex>
+#include <shared_mutex>
 
 #ifdef _MSC_VER
 // Needed for Visual Studio to disable runtime checks for mempcy
@@ -814,32 +814,13 @@ struct Manhattan : Minkowski {
   }
 };
 
-class ThreadBarrier {
-public:
-  explicit ThreadBarrier(std::size_t count) : _count(count) { }
-
-  void wait() {
-    std::unique_lock<std::mutex> lock(_mutex);
-    if (--_count == 0) {
-      _cv.notify_all();
-    } else {
-      _cv.wait(lock, [this] { return _count == 0; });
-    }
-  }
-
-private:
-  std::mutex _mutex;
-  std::condition_variable _cv;
-  std::size_t _count;
-};
-
 template<typename S, typename T>
 class AnnoyIndexInterface {
  public:
   // Note that the methods with an **error argument will allocate memory and write the pointer to that string if error is non-NULL
   virtual ~AnnoyIndexInterface() {};
   virtual bool add_item(S item, const T* w, char** error=NULL) = 0;
-  virtual bool build(int q, int n_threads=1, char** error=NULL) = 0;
+  virtual bool build(int q, int n_threads=-1, char** error=NULL) = 0;
   virtual bool unbuild(char** error=NULL) = 0;
   virtual bool save(const char* filename, bool prefault=false, char** error=NULL) = 0;
   virtual void unload() = 0;
@@ -952,7 +933,7 @@ public:
     return true;
   }
     
-  bool build(int q, int n_threads=1, char** error=NULL) {
+  bool build(int q, int n_threads=-1, char** error=NULL) {
     if (_loaded) {
       set_error_from_string(error, "You can't build a loaded index");
       return false;
@@ -963,21 +944,33 @@ public:
       return false;
     }
 
-    if (n_threads < 1) {
-      set_error_from_string(error, "You can't build an index with less than 1 thread");
-      return false;
+    if (n_threads == -1) {
+      // If the hardware_concurrency() value is not well defined or not computable, it returns ​0.
+      // We guard against this by using at least 1 thread.
+      n_threads = std::max(1, (int)std::thread::hardware_concurrency());
     }
 
     D::template preprocess<T, S, Node>(_nodes, _s, _n_items, _f);
 
     _n_nodes = _n_items;
 
-    std::mutex nodes_mutex;
-    ThreadBarrier barrier(n_threads);
+    std::shared_timed_mutex nodes_mutex;
+    std::mutex n_nodes_mutex;
+    std::mutex roots_mutex;
+    std::mutex nodes_size_mutex;
     vector<std::thread> threads(n_threads);
-    for (int i = 0; i < n_threads; i++) {
-      int trees_per_thread = q == -1 ? -1 : (int)floor((q + i) / n_threads);
-      threads[i] = std::thread(&AnnoyIndex<S, T, D, Random>::_thread_build, this, trees_per_thread, i, std::ref(barrier), std::ref(nodes_mutex));
+    for (int thread_idx = 0; thread_idx < n_threads; thread_idx++) {
+      int trees_per_thread = q == -1 ? -1 : (int)floor((q + thread_idx) / n_threads);
+      threads[thread_idx] = std::thread(
+        &AnnoyIndex<S, T, D, Random>::_thread_build,
+        this,
+        trees_per_thread,
+        thread_idx,
+        std::ref(nodes_mutex),
+        std::ref(n_nodes_mutex),
+        std::ref(nodes_size_mutex),
+        std::ref(roots_mutex)
+      );
     }
 
     for (auto& thread : threads) {
@@ -1173,49 +1166,61 @@ public:
   }
 
 protected:
+  void _reallocate_nodes(S n) {
+    const double reallocation_factor = 1.3;
+    S new_nodes_size = std::max(n, (S) ((_nodes_size + 1) * reallocation_factor));
+    void *old = _nodes;
+    
+    if (_on_disk) {
+      if (!remap_memory_and_truncate(&_nodes, _fd, 
+          static_cast<size_t>(_s) * static_cast<size_t>(_nodes_size), 
+          static_cast<size_t>(_s) * static_cast<size_t>(new_nodes_size)) && 
+          _verbose)
+          showUpdate("File truncation error\n");
+    } else {
+      _nodes = realloc(_nodes, _s * new_nodes_size);
+      memset((char *) _nodes + (_nodes_size * _s) / sizeof(char), 0, (new_nodes_size - _nodes_size) * _s);
+    }
+    
+    _nodes_size = new_nodes_size;
+    if (_verbose) showUpdate("Reallocating to %d nodes: old_address=%p, new_address=%p\n", new_nodes_size, old, _nodes);
+  }
+
   void _allocate_size(S n) {
     if (n > _nodes_size) {
-      const double reallocation_factor = 1.3;
-      S new_nodes_size = std::max(n, (S) ((_nodes_size + 1) * reallocation_factor));
-      void *old = _nodes;
-      
-      if (_on_disk) {
-        if (!remap_memory_and_truncate(&_nodes, _fd, 
-            static_cast<size_t>(_s) * static_cast<size_t>(_nodes_size), 
-            static_cast<size_t>(_s) * static_cast<size_t>(new_nodes_size)) && 
-            _verbose)
-            showUpdate("File truncation error\n");
-      } else {
-        _nodes = realloc(_nodes, _s * new_nodes_size);
-        memset((char *) _nodes + (_nodes_size * _s) / sizeof(char), 0, (new_nodes_size - _nodes_size) * _s);
-      }
-      
-      _nodes_size = new_nodes_size;
-      if (_verbose) showUpdate("Reallocating to %d nodes: old_address=%p, new_address=%p\n", new_nodes_size, old, _nodes);
+      _reallocate_nodes(n);
     }
+  }
+
+  void _allocate_size(S n, std::shared_timed_mutex& nodes_mutex, std::mutex& nodes_size_mutex) {
+    nodes_size_mutex.lock();
+    if (n > _nodes_size) {
+      nodes_mutex.lock();
+      _reallocate_nodes(n);
+      nodes_mutex.unlock();
+    }
+    nodes_size_mutex.unlock();
   }
 
   inline Node* _get(const S i) const {
     return get_node_ptr<S, Node>(_nodes, _s, i);
   }
 
-  void _thread_build(int q, int thread_idx, ThreadBarrier& barrier, std::mutex& nodes_mutex) {
+  void _thread_build(int q, int thread_idx, std::shared_timed_mutex& nodes_mutex, std::mutex& n_nodes_mutex, std::mutex& nodes_size_mutex, std::mutex& roots_mutex) {
     Random _random;
     // Each thread needs its own seed, otherwise each thread would be building the same tree(s)
     int seed = _is_seeded ? _seed + thread_idx : thread_idx;
     _random.set_seed(seed);
 
-    vector<vector<Node*> > thread_trees;
     vector<S> thread_roots;
     while (1) {
       if (q == -1) {
-        size_t thread_n_nodes = 0;
-        for (size_t tree_idx = 0; tree_idx < thread_trees.size(); tree_idx++) {
-          thread_n_nodes += thread_trees[tree_idx].size();
-        }
-        if (thread_n_nodes >= 2 * (size_t)_n_items) {
+        n_nodes_mutex.lock();
+        if (_n_nodes >= 2 * _n_items) {
+          n_nodes_mutex.unlock();
           break;
         }
+        n_nodes_mutex.unlock();
       } else {
         if (thread_roots.size() >= (size_t)q) {
           break;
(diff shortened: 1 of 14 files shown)


novoselrok:
Summary of CI issues:
* Missing a cross-platform implementation of isnan
* Missing -std=c++11 flag when compiling Lua and Go extensions
* Windows Python 2.7 cannot find mutex library

src/annoylib.h
```diff
@@ -940,23 +963,32 @@ template<typename S, typename T, typename Distance, typename Random>
...
-          indices.push_back(i);
+    std::mutex _nodes_mutex;
+    ThreadBarrier barrier(n_threads);
+    vector<std::thread> threads(n_threads);
+    int work_per_thread = (int)floor(q / (double)n_threads);
+    int work_remainder = q % n_threads;
+    for (int i = 0; i < n_threads; i++) {
+      int trees_per_thread = -1;
+      if (q > -1) {
+        // First thread picks up the remainder of the work
```
erikbern, line 983: this might lead to an uneven split, eg if `q=20` and `n_threads=7` then the distribution will be `[8, 2, 2, 2, 2, 2, 2]`

I think you can do `trees_per_thread  work_per_thread + (i < work_remainder)` and it should split it like `[3, 3, 3, 3, 3, 3, 2]`

src/annoylib.h
```diff
@@ -1172,7 +1206,87 @@ template<typename S, typename T, typename Distance, typename Random>
     return get_node_ptr<S, Node>(_nodes, _s, i);
   }
 
-  S _make_tree(const vector<S >& indices, bool is_root) {
+  void _thread_build(int q, int thread_idx, ThreadBarrier& barrier, std::mutex& _nodes_mutex) {
```
erikbern, line 1209: i think it would be good to have the same code for threaded and non-threaded building

src/annoylib.h
```diff
@@ -940,23 +963,32 @@ template<typename S, typename T, typename Distance, typename Random>
...
-      if (_verbose) showUpdate("pass %zd...\n", _roots.size());
 
-      vector<S> indices;
-      for (S i = 0; i < _n_items; i++) {
-        if (_get(i)->n_descendants >= 1) // Issue #223
-          indices.push_back(i);
+    std::mutex _nodes_mutex;
+    ThreadBarrier barrier(n_threads);
+    vector<std::thread> threads(n_threads);
+    int work_per_thread = (int)floor(q / (double)n_threads);
```
erikbern, line 978: what happens if `q = -1`? you probably have to determine `q` before this code in case it's set to `-1`

erikbern:
this is super cool! i left a few minor comment, but generally this is great

src/annoylib.h
```diff
@@ -929,7 +952,7 @@ template<typename S, typename T, typename Distance, typename Random>
     return true;
   }
     
-  bool build(int q, char** error=NULL) {
+  bool build(int q, int n_threads=1, char** error=NULL) {
```
erikbern, line 955: might be good to default it to `-1` and if that's the case set it to the number of cpus

src/annoylib.h
```diff
@@ -1172,7 +1206,87 @@ template<typename S, typename T, typename Distance, typename Random>
...
+  void _thread_build(int q, int thread_idx, ThreadBarrier& barrier, std::mutex& _nodes_mutex) {
+    Random _random;
+    // Each thread needs its own seed, otherwise each thread would be building the same tree(s)
+    int seed = _is_seeded ? _seed + thread_idx : thread_idx;
+    _random.set_seed(seed);
+
+    vector<vector<Node*> > thread_trees;
+    vector<S> thread_roots;
+    while (1) {
+      if (q == -1) {
```
erikbern, line 1218: this code to determine `q` if `q = -1` could probably just be left in the `build` function

src/annoylib.h
erikbern: actually even simpler you can just to `trees_per_thread = (q + i) / n_threads` and you don't even have to define `work_pre_thread`

novoselrok: Awesome, I was not exactly happy with my solution, so I'll reuse this.

src/annoylib.h
novoselrok: I'm not exactly sure what you mean by this: should we still support building without threads (e.g. when n_threads=1)?

erikbern: I'd actually be ok only supporting threaded build so that we can remove the non-threaded build code

src/annoylib.h
novoselrok: The thing with `q = -1`, is that I don't know beforehand how many trees each thread should build, because it depends on the number of nodes. That is why I have to check for `-1` everywhere.

A better solution would be to introduce a different meaning to `q = -1`, where each thread would build just one tree for example.

src/annoylib.h
novoselrok: I'm probably missing something, but `_thread_build` is just a private helper function that was extracted from the previous `build` function. It is there just so I don't have to stuck everything inside a lambda function when spawning a new thread.

Could you point out which part of the code could be potentially removed? :smile:

novoselrok:
@erikbern Could you take a look at the failing Go and Lua builds, if you'll be able to help? I'm not sure how to get the -std=c++11 flag into the compilation.

Windows Python 2.7 might be tricky, because it seems it doesn't support C++11 thread/mutex at all.

src/annoylib.h
erikbern: Ah yeah good point. I forgot about that. Ok if your code works then let's keep it, I guess

src/annoylib.h
```diff
@@ -1172,7 +1199,87 @@ template<typename S, typename T, typename Distance, typename Random>
...
+      // right after the already inserted items. Indices of split nodes start with _n_items, n_items + 1, ...
+      // We do not want to mutate the _nodes array during tree construction due to reallocation issues. That is
+      // why each thread stores the trees locally until all threads are ready to insert them into _nodes.
+      S root_node = _make_tree(indices, split_nodes, true, _random);
+      thread_roots.push_back(root_node);
+      thread_trees.push_back(split_nodes);
+    }
+
+    // Wait for all threads to finish before we can start inserting tree nodes into global _nodes array
+    barrier.wait();
```
erikbern, line 1244: isn't locking the mutex enough to enforce this?

src/annoylib.h
erikbern: I don't really understand why `_thread_build` is even needed? Couldn't you just call `_make_tree` and make sure each thread locks the part of the code where nodes are added? The onlyt two places are 
* https://github.com/spotify/annoy/blob/master/src/annoylib.h#L1185
* https://github.com/spotify/annoy/blob/master/src/annoylib.h#L1252

I don't think you need the magic where you go into nodes and add offsets etc? I could be wrong

erikbern: If it's not clear what I'm saying then give me a few days and I can try to get it working and then send a pull request to your pull request!

novoselrok: That is how I did it in my initial implementation, but the problem lies in the `_allocate_size` function.

When multiple threads are building trees and one of them has to resize the memory (with `realloc`), it invalidates all pointers  held by other threads here: https://github.com/spotify/annoy/blob/master/src/annoylib.h#L1199
This leads to a segmentation fault. To guard against this, each thread would have to copy the children before using them. Additionally, all `_allocate_size` calls would have to be guarded by an exclusive mutex and all `_get` calls with a shared mutex (because we would have to ensure `_get` does not return an invalid pointer). This resulted in a significant slow-down due to mutex contention and increase in memory usage due to copying.

src/annoylib.h
novoselrok: Unfortunately not, due to reasons noted here: https://github.com/spotify/annoy/pull/495#discussion_r461732904 We have to wait for all threads finish their work before we can mutate the global _nodes buffer.

src/annoymodule.cc
```diff
@@ -408,16 +408,17 @@ py_an_on_disk_build(py_annoy *self, PyObject *args, PyObject *kwargs) {
 static PyObject *
 py_an_build(py_annoy *self, PyObject *args, PyObject *kwargs) {
   int q;
+  int n_threads = 1;
   if (!self->ptr) 
     return NULL;
-  static char const * kwlist[] = {"n_trees", NULL};
-  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "i", (char**)kwlist, &q))
+  static char const * kwlist[] = {"n_trees", "n_threads", NULL};
```
pavlin-policar, line 414: Just an aside -- what do you think about naming this parameter in the Python wrapper to `n_jobs` instead of `n_threads`? `n_jobs` seems more consistent with the Python ecosystem. Either way, I think this is a great improvement to Annoy!

src/annoylib.h
erikbern: Ok, that makes a lot of sense!

Instead of pointers to nodes, you could store their indices in the `children` array. That way the pointers won't get invalidated. However like you said, `_get` would still need to acquire a shared mutex. But I think instead of putting the mutex on `_get`, if you use your barrier approach, then aren't we guaranteed no thread will update the underlying node storage while it's being accessed? Not 100%

src/annoymodule.cc
novoselrok: Sure, that makes sense. I can expose `n_jobs` keyword argument for Python, but keep `n_threads` everywhere else.

src/annoylib.h
novoselrok: Depends on where you would put the barrier. It sounds like it might work, but I'm worried this might significantly impact performance, because `_nodes` array is constantly accessed/written to. That is why I wanted to get rid of any critical sections in `_make_tree`, to make threads as independent as possible.

erikbern: When you said putting a mutex on `_get` resulted in a significant slow-down, was that during the tree building phase? Or during the search phase?

novoselrok: Just the building phase. I put the lock around the parts you linked to (including the `_get` call).

novoselrok: Another tricky thing might be support for `std::shared_mutex` (provides shared and exclusive locking capabilities), which is only supported from C++17 onward.

Other than that, I think your proposal could work. `create_split` would have to take the `_nodes` array as a parameter and a read lock would have to be placed around `create_split` and around the parts where `Node` pointers are active (returned from `_get`). And `_allocate_size` would have to protected with an exclusive lock.

erikbern: Yeah exactly, that's what i was thinking.

Your approach at https://github.com/spotify/annoy/pull/495/files#diff-91eaadc9d0aa248375a22a64f315e3d8R1250 is very clever but i'm a bit nervous it's too hacky and I'd love to have something slightly simpler :)

Doesn't your barrier thing roughly accomplish what at shared_mutex does? Or it should at least be possible to modify it into sort of a shared mutex? Haven't thought about it enough.

novoselrok: I agree, I would definitely feel more comfortable if I could simplify it a bit :smile: 

Unfortunately a barrier does not accomplish what a shared_mutex is supposed to do. It can be only used as a global synchronization point, not for reader/writer style locking.

After reading a couple of SO threads, a custom shared mutex could prove really tricky to implement correctly. We could use the `shared_timed_mutex` which has a few additional capabilities and is available from C++14. Either way, I'll implement the less-magic version we discussed and then we can compare the performance and decide if it's worth it to bump the C++ version.

novoselrok: Well, I managed to simplify my solution in the latest commit :smile: I removed all of the magic and introduced a couple of more `mutex` variables and a `shared_timed_mutex` variable.

I've re-ran my time measurements on glove-angular-25:
| Num. threads | n_trees=32 time (s) | n_trees=64 time (s) |
|--------------|---------------------|---------------------|
|            1 |  49.286592960357666 | 99.88854050636292   |
|            2 |   26.54487133026123 | 53.29321575164795   |
|            4 |  15.026970863342285 | 29.342100143432617  |
|            8 |  10.422403812408447 | 18.966835498809814  |
|           16 |   9.757703065872192 | 17.069363117218018  |

erikbern: Wow amazing! Any idea if there's any slowdown compared to the current master?

The new code looks much cleaner and less "clever". I don't know how well supported c++14 is but hopefully it's not a big problem.

Do you really need so many different mutexes? Not a concern, was just surprised to see 4 different ones

novoselrok: I can prepare a more comprehensive timing benchmark and include a comparison with the current master. But as far as I can tell, there isn't a significant slowdown.

Regarding the number of mutexes: If I guard everything with the shared `nodes_mutex` just one thread has to acquire an exclusive lock and everything grinds to a halt. So I opted for a more granular approach, where I exclusively lock the `nodes_mutex` only when absolutely necessary and add a separate regular locks for all other shared variables (`nodes_size`, `_n_nodes`). I'll see if I can cut down on the number of mutexes, but I currently can't see a better way.

erikbern:
This looks quite amazing so far. Would be good to figure out the CI failures

novoselrok:
> This looks quite amazing so far. Would be good to figure out the CI failures

As far as I can tell Windows Py2.7 does not support the C++ thread/mutex libraries in any way. For Go and Lua I need to add the -std=c++14 flag to the compilation step, but I'm not really sure how to do that.

erikbern:
maybe there's some dumb way so that if those headers aren't present, it just defaults to single-threaded? not sure if that's a lot of work. i'd imagine there would be a lot of `#ifdef` scattered throughout the code

erikbern:
FYI the tests timed out on OSX which implies there might be some race condition or similar? https://travis-ci.org/github/spotify/annoy/jobs/713675853

erikbern:
nice, the last change is 👌

novoselrok:
> nice, the last change is ok_hand

Feel free to take another look and see if anything can be improved. OSX is going wonky, I'm still investigating the issue.

novoselrok:
I think I figured out the OSX issue. Apparently locking/unlocking is much more expensive under OSX and I'm doing it excessively in a loop when adding indices. Moving the lock around the loop should do the trick 🤞

erikbern:
this looks great – i'm happy to merge whenever you give 👍

novoselrok:
Awesome 👍  I'll just prepare a short timing comparison with master sometime tomorrow. If there aren't any more surprises, it'll be ready to go.

novoselrok:
Here is a multihreaded build timing comparison for angular-glove-25 dataset:

| Num. threads | n_trees=16 time (s) | n_trees=32 time (s) | n_trees=64 time (s) | n_trees=128 time (s) |
|--------------|---------------------|---------------------|---------------------|----------------------|
|            1 |         24.35913038 |          49.1573956 |          98.2360599 |          197.1160657 |
|            2 |         13.40727019 |         25.64709997 |         51.36644101 |          104.5556092 |
|            4 |         7.058898449 |         14.07375836 |         29.17391777 |          58.58751988 |
|            8 |         4.247594595 |         8.404431581 |         17.18267322 |          33.64984083 |
|           16 |         3.579955816 |         7.196531773 |         14.31243014 |          28.61186814 |

Here are single threaded timings from master for comparison:
| Num. trees | Time (s)           |
|------------|--------------------|
| 16         | 28.233051300048828 |
| 32         | 53.763163566589355 |
| 64         | 102.16749143600464 |
| 128        | 212.41096830368042 |

The speed up is not linear, but it's good enough especially for large number of trees.

erikbern:
Awesome, I’ll merge this now.

I won’t publish this to pypi for a few weeks just in case there’s some issues. There’s some people using master so that’s effectively a form of “beta testing”

Really appreciate the work!

novoselrok:
Awesome, sounds good! :smile:

n1ru4l:
Hey there, we are already using the multithreaded build with python and love the speed improvements! Now we are thinking even thinking about migrating to go in order to have better performance for loading the data before inserting it into the tree.

> As far as I can tell Windows Py2.7 does not support the C++ thread/mutex libraries in any way. For Go and Lua I need to add the -std=c++14 flag to the compilation step, but I'm not really sure how to do that.

Is there a specific issue that blocked this from getting implemented? Would you plan to follow-up with a PR that enabled multithreaded builds, or should I try sending a PR?

novoselrok:
Hey @n1ru4l, no I wasn't planning on it. You can give it a go, if you would like :) As I've mentioned, you would only need to supply the -std=c++14 flag when compiling for Go and possibly update the Go version in tests (https://github.com/spotify/annoy/blob/master/tox.ini#L13).

n1ru4l:
@novoselrok That sounds easy in theory :D I will give it a try!

n1ru4l:
So we decided against using golang. I won't implement this any time soon then.

utkarshgupta137:
I'm running this in docker on ubuntu, but I'm only seeing 1 CPU core used at a time (100%), the rest are idle (0%).
I'm passing n_jobs = -1 & also tried to set it to 6 (which is the number of cores), but got the same result.

When I'm training a tensorflow model in the same docker, it uses 100% of the CPUs, so I don't think docker configuration is the issue.

AniruddhaHumane:
I am facing same issue as @utkarshgupta137, did you find any solution for this?

utkarshgupta137:
> I am facing same issue as @utkarshgupta137, did you find any solution for this?

It has been a while, but I don't think I did.

AniruddhaHumane:
I see. That's unfortunate. I have created an issue here -> https://github.com/spotify/annoy/issues/627 hopefully the author will respond.

Issue #382, opened 2019-04-10
search_k documentation is inconsistent


txntxn:
The documentation for `search_k` is inconsistent.

In some places, it says: "search_k nodes which defaults to n_trees * n if not provided"

In other places, it says: "If search_k is not provided, it will default to n * n_trees * D where n is the number of approximate nearest neighbors and D is a constant depending on the metric."

Which is it?

And if it is `D`, could you enumerate the different `D` values for each metric? I searched the code for a few minutes but I was unable to determine what D is.

erikbern:
where does it say something about D? that seems incorrect

txntxn:
In the README. Perhaps in other places too? But definitely in the README.

erikbern:
ok will take a look and fix

PR #497, opened 2020-08-05
Remove D from search_k documentation

novoselrok:
Issue: #382

Removed D from search_k tradeoffs documentation, because it is not present in the code.


erikbern:
Thanks! I don't even remember what that was for

novoselrok:
No problem, you can now probably close https://github.com/spotify/annoy/issues/382 🙂

PR #500, opened 2020-08-10
Add link to Ruby bindings for Annoy

yoshoku:
Thank you for the awesome library. There are [Ruby bindings for Annoy](https://github.com/yoshoku/annoy.rb). I would be grateful if you could add it to the list on the More info section.


erikbern:
Nice thanks!

PR #509, opened 2020-09-30
Add dot product to gowrapper

fredrikluo:
Add the necessary classes so golang can use the dot product distance.


erikbern:
nice!

PR #525, opened 2020-12-25
unuse AVX512 build option add

aie2:
I want to build annoy without AVX512.
if env variable UNUSE_AVX512_ANNOY_COMPILE = YES, it builds without AVX512.


setup.py
```diff
@@ -32,6 +33,29 @@
 
 """
 
+def _get_AVX_macro_list():
+    cmd = """gcc -march=native -dM -E - < /dev/null | grep "__AVX512" | sort"""
```
erikbern, line 37: what does this command do?

aie2: The part of「gcc -march=native -dM -E - < /dev/null」
generates a list of ‘#define’ directives for all the macros defined during the execution of the preprocessor.
  c.f. https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#Preprocessor-Options

And, 「grep "__AVX512"」 extracts AVX512 macro.

erikbern: this feels a bit hacky – i'm nervous this won't work on all platforms (windows definitely not)

aie2: So, that's it. 
In windows gcc doesn't exists, so this code can't work.
sorry, I rethink.

erikbern:
I'm hoping to add functionality to detect AVX capabilities in runtime at some point. I think that would be a much better way to disable it if the host process doesn't support it. So this might not be needed if I do that?

aie2:
If you do that, I'm glad. Well then, I close this PR.

PR #529, opened 2021-01-20
get rid of nose

MeggyCal:
As nose is unmaintained for a while, let us replace it with unittest.


MeggyCal:
... and tests don't work, because I don't know how to setup your environment :( would you help me with that, please?

erikbern:
Thanks for the PR! I think it fails because of something with how `python setup.py nosetests` builds the library before it runs the test. You might have to add a separate step to build the module first. I can clone and take a look.

nadzhou:
> Thanks for the PR! I think it fails because of something with how `python setup.py nosetests` builds the library before it runs the test. You might have to add a separate step to build the module first. I can clone and take a look.

Should be easy to do I'd think. A simple `pip install -e .` or adding the repo directory to $PATH.

PR #548, opened 2021-05-13
add type annotations

### LINKED ISSUE #545: Python type annotations
Hi,

Would you accept a pull request to add type annotations? If not, would you be okay with type annotations being added for annoy in http://github.com/python/typeshed?

Also, if type annotations were added, would you want to also type check the annoy tests?

chdsbd:
fixes #545


erikbern:
why are the type annotations in a different file and not in the main .py file?

chdsbd:
To add type annotations for the Annoy C module we need to use a separate [stubs file](https://mypy.readthedocs.io/en/stable/stubs.html). These stub files end with `.pyi` and type checkers like [mypy](http://www.mypy-lang.org) will use them for typing the package.

If Annoy was plain Python code we could add the type annotations inline.

On the flip side, by keeping the annotations in a separate stub file, we don't need to deprecate older versions of Python.

erikbern:
cool, let me know when this is ready to be merged!

chdsbd:
@erikbern This change is ready to merge. thanks!

PR #550, opened 2021-05-14
improve error handling for vectors

### LINKED ISSUE #531: `ValueError` showing incorrect message
Hello,

I was trying to use annoy with my data and got an error with `t.add_item(m, data[m,:])`.
The error points out that `data[m,:]` is not an iterable, but it is indeed an iterable object.

#### Minimal Working example:

```python
import scipy

from collections.abc import Iterable

def check_iterable(input_vec):
    if isinstance(input_vec, Iterable):
        print(f"Input {type(input_vec)} it is iterable")

X_dense = np.random.rand(100,1000)        
X_sp = scipy.sparse.csr_matrix(X_dense)

check_iterable(X_dense)
check_iterable(X_sp)

def build_annoy_index(data, n_trees):
    print(f"Building annoy index with data type {type(data)}")
    n_examples, n_features = data.shape
    t = AnnoyIndex(n_features, 'angular')  
    
    for m,x in enumerate(range(n_examples)):
        t.add_item(m, data[m,:])

    t.build(n_trees = n_trees) # 10 trees
    t.save(f'AIndex_Nexamples={n_examples}=_Ntrees={n_trees}.ann')


n_trees = 10
build_annoy_index(X_dense[0:20], n_trees)
build_annoy_index(X_sp[0:20], n_trees)
```

outputs the following:
```python
Input <class 'numpy.ndarray'> it is iterable
Input <class 'scipy.sparse.csr.csr_matrix'> it is iterable
Building annoy index with data type <class 'numpy.ndarray'>
Building annoy index with data type <class 'scipy.sparse.csr.csr_matrix'>
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-108-51459ca9a2d5> in <module>
     27 n_trees = 10
     28 build_annoy_index(X_dense[0:20], n_trees)
---> 29 build_annoy_index(X_sp[0:20], n_trees)

<ipython-input-108-51459ca9a2d5> in build_annoy_index(data, n_trees)
     19 
     20     for m,x in enumerate(range(n_examples)):
---> 21         t.add_item(m, data[m,:])
     22 
     23     t.build(n_trees = n_trees) # 10 trees

ValueError: Expected an iterable, got an object of type "csr_matrix"
```

### LINKED ISSUE #546: Missing error checks in Python extension code
The code in https://github.com/spotify/annoy/blob/master/src/annoymodule.cc that interacts with the CPython API mostly doesn't check the return value of called functions. This can lead to crashes:

```
% python -c 'import annoy
class X:
    def __len__(self): return 42
ann = annoy.AnnoyIndex(42)
ann.get_nns_by_vector(X(), 42)
'
<string>:4: FutureWarning: The default argument for metric will be removed in future version of Annoy. Please pass metric='angular' explicitly.
zsh: segmentation fault  python -c 
```
Similarly, calling `get_nns_by_vector()` with a vector of objects that aren't floats throws a `SystemError`.

There are also no checks on the return value of functions like `PyList_New`. In theory, this could lead to segfaults, because the function returns NULL on error. I believe that only happens if the system is out of memory though.

chdsbd:
Annoy now better handles invalid vectors.

We now use Python's default error formatting when a vector doesn't have a `__len__` method. We also check the return types for functions like `PyObject_GetItem` and `PyFloat_AsDouble` so we gracefully handle errors.

fixes #531
fixes #546

### DIFF
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -301,7 +301,7 @@ py_an_get_nns_by_item(py_annoy *self, PyObject *args, PyObject *kwargs) {
 
 
 bool
-convert_iterable_to_vector(PyObject* v, int f, vector<float>* w) {
+convert_list_to_vector(PyObject* v, int f, vector<float>* w) {
   Py_ssize_t length = PyObject_Size(v);
   if (length == -1) {
     return false;
@@ -311,30 +311,14 @@ convert_iterable_to_vector(PyObject* v, int f, vector<float>* w) {
     return false;
   }
 
-  PyObject *iterator = PyObject_GetIter(v);
-  PyObject *item;
-
-  if (iterator == NULL) {
     return false;
   }
   for (int z = 0; z < f; z++) {
-    item = PyIter_Next(iterator);
-
-    if (item == NULL) {
-      PyErr_Format(PyExc_IndexError, "Vector has wrong length (expected %d, got %d)", f, z);
-      return false;
-    }
-
-    double value = PyFloat_AsDouble(item);
-    Py_DECREF(item);
-    if (value == -1.0 && PyErr_Occurred()) {
-      return false;
-    }
-    (*w)[z] = value;
-  }
-  Py_DECREF(iterator);
-  if (PyErr_Occurred()) {
-    return false;
+    PyObject *key = PyInt_FromLong(z);
+    PyObject *pf = PyObject_GetItem(v, key);
+    (*w)[z] = PyFloat_AsDouble(pf);
+    Py_DECREF(key);
+    Py_DECREF(pf);
   }
   return true;
 }
@@ -351,7 +335,7 @@ py_an_get_nns_by_vector(py_annoy *self, PyObject *args, PyObject *kwargs) {
     return NULL;
 
   vector<float> w(self->f);
-  if (!convert_iterable_to_vector(v, self->f, &w)) {
+  if (!convert_list_to_vector(v, self->f, &w)) {
     return NULL;
   }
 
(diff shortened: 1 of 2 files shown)


test/types_test.py
```diff
@@ -56,3 +56,53 @@ def test_range_errors(self, n_points=1000, n_trees=10):
...
+        """
+        class FakeCollection:
+            def __len__(self):
+                return 3
+            
+            def __getitem__(self, i):
+                raise IndexError 
+        
+        i = AnnoyIndex(3, 'euclidean')
+        with self.assertRaises(IndexError):
```
erikbern, line 97: looks like this raises an AttributeError on python 2.7, is it easy to fix? otherwise i'm maybe fine dropping support for 2.7 soon anyway...

erikbern:
super nice!

chdsbd:
@erikbern I added a version check for Python 2.7 so now the tests should pass on 2.7.

erikbern:
Thanks @chdsbd !!!

src/annoymodule.cc
```diff
@@ -237,16 +237,25 @@ py_an_save(py_annoy *self, PyObject *args, PyObject *kwargs) {
...
+    return NULL;
+  }
   for (size_t i = 0; i < result.size(); i++)
     PyList_SetItem(l, i, PyInt_FromLong(result[i]));
   if (!include_distances)
     return l;
 
   PyObject* d = PyList_New(distances.size());
+  if (d == NULL) {
+    return NULL;
```
JelleZijlstra, line 250: This leaks a reference to `l`. The standard approach is to have a `goto error`, and the `error` block `Py_XDECREF`s all the things that may have been allocated.

chdsbd: I've opened a PR to address your feedback. if you don't mind it would be great to have your review

https://github.com/spotify/annoy/pull/552

PR #552, opened 2021-05-19
fix leaking references in error handling

chdsbd:
this change addresses feedback from: https://github.com/spotify/annoy/pull/550/files#r635611735

We need to ensure we decrement our references on error to prevent leaking references.

### DIFF
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -267,6 +267,7 @@ get_nns_to_python(const vector<int32_t>& result, const vector<float>& distances,
     if (PyList_SetItem(d, i, dist) < 0) {
       goto error;
     }
+    dist = NULL;
   }
 
   if ((t = PyTuple_Pack(2, l, d)) == NULL) {
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -250,6 +250,7 @@ get_nns_to_python(const vector<int32_t>& result, const vector<float>& distances,
       goto error;
     }
     PyList_SetItem(l, i, res);
+    res = NULL;
   }
   if (!include_distances)
     return l;
@@ -263,6 +264,7 @@ get_nns_to_python(const vector<int32_t>& result, const vector<float>& distances,
       goto error;
     }
     PyList_SetItem(d, i, dist);
+    dist = NULL;
   }
 
   if ((t = PyTuple_Pack(2, l, d)) == NULL) {
@@ -400,6 +402,7 @@ py_an_get_item_vector(py_annoy *self, PyObject *args) {
       goto error;
     }
     PyList_SetItem(l, z, dist);
+    dist = NULL;
   }
 
   return l;
--- a/src/annoymodule.cc
+++ b/src/annoymodule.cc
@@ -239,14 +239,13 @@ get_nns_to_python(const vector<int32_t>& result, const vector<float>& distances,
   PyObject* l = NULL;
   PyObject* d = NULL;
   PyObject* t = NULL;
-  PyObject* res = NULL;
-  PyObject* dist = NULL;
 
   if ((l = PyList_New(result.size())) == NULL) {
     goto error;
   }
   for (size_t i = 0; i < result.size(); i++) {
-    if ((res = PyInt_FromLong(result[i])) == NULL) {
+    PyObject* res = PyInt_FromLong(result[i]);
+    if (res == NULL) {
       goto error;
     }
     if (PyList_SetItem(l, i, res) < 0) {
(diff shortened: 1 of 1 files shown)


src/annoymodule.cc
```diff
@@ -236,30 +236,37 @@ py_an_save(py_annoy *self, PyObject *args, PyObject *kwargs) {
...
+  PyObject* l = NULL;
+  PyObject* d = NULL;
+  PyObject* t = NULL;
+  l = PyList_New(result.size());
   if (l == NULL) {
-    return NULL;
+    goto error;
   }
   for (size_t i = 0; i < result.size(); i++)
     PyList_SetItem(l, i, PyInt_FromLong(result[i]));
```
JelleZijlstra, line 247: Not sure what's going on with `PyInt` vs. `PyLong`, but this can also fail: https://docs.python.org/3/c-api/long.html#c.PyLong_FromLong

`PyList_SetItem` returns -1 on failure, but I believe that's only if the index is out of bounds, which should never happen.

src/annoymodule.cc
```diff
@@ -236,30 +236,37 @@ py_an_save(py_annoy *self, PyObject *args, PyObject *kwargs) {
...
     return l;
 
-  PyObject* d = PyList_New(distances.size());
+  d = PyList_New(distances.size());
   if (d == NULL) {
-    return NULL;
+    goto error;
   }
   for (size_t i = 0; i < distances.size(); i++)
     PyList_SetItem(d, i, PyFloat_FromDouble(distances[i]));
```
JelleZijlstra, line 256: `PyFloat_FromDouble` needs an error check: https://docs.python.org/3/c-api/float.html#c.PyFloat_FromDouble

JelleZijlstra (commented):
Thanks! There's a few more missing checks I noticed.

I can't comment outside the part you changed, but I also looked at the rest of the file, and `py_an_get_item_vector` has another `PyFloat_FromDouble` call that is not checked.

erikbern:
Looks good but is this actually a problem in reality? Does PyList_New ever fail other than in very strange situations (like you're about to get an OOM anyway)?

JelleZijlstra:
Yes, `PyLIst_New` basically only fails if memory allocation fails. In the environments I'd be likely to run this code that's unlikely to ever happen (the Linux OOM killer would kill the process first), but that may not be true for all users.

erikbern:
i guess it doesn't hurt... is this ready to be merged?

chdsbd:
I made some more changes in response to @JelleZijlstra 's feedback

src/annoymodule.cc
```diff
@@ -236,30 +236,52 @@ py_an_save(py_annoy *self, PyObject *args, PyObject *kwargs) {
...
-  PyTuple_SetItem(t, 0, l);
-  PyTuple_SetItem(t, 1, d);
 
   return t;
+
+  error:
+    Py_XDECREF(l);
+    Py_XDECREF(d);
+    Py_XDECREF(t);
+    Py_XDECREF(res);
```
JelleZijlstra, line 282: This is actually incorrect because PyList_SetItem steals a reference to `res`. So if we get here because `d = PyList_New` failed, we'll decref the last element of the `l` list twice: once when the list itself gets deleted, and once from this line. I believe setting to NULL at the end of the loop body will fix this.

Similar reasoning applies to the `dist` variable in the second for loop.

chdsbd: fixed. Thanks for the feedback!

JelleZijlstra: Looks like you fixed only the second instance (dist), not the first one (res)

chdsbd: I think I've fixed this now with my latest commit

chdsbd: nvm, I messed it up again 🤦

src/annoymodule.cc
```diff
@@ -236,30 +236,50 @@ py_an_save(py_annoy *self, PyObject *args, PyObject *kwargs) {
...
+  PyObject* dist = NULL;
+
+  if ((l = PyList_New(result.size())) == NULL) {
+    goto error;
+  }
+  for (size_t i = 0; i < result.size(); i++) {
+    if ((res = PyInt_FromLong(result[i])) == NULL) {
+      goto error;
+    }
+    PyList_SetItem(l, i, res);
```
chdsbd, line 252: Our calls to PyList_SetItem will always succeed

src/annoymodule.cc
chdsbd: Because of the way we call PyList_SetItem, it will always succeed. We don't need to check the error response.

I _think_ I've actually resolved the double decref's.

erikbern:
let me know when this is ready

JelleZijlstra (approved):

PR #562, opened 2021-07-09
Support CMake-based dependency resolution in C++ projects

LTLA:
This PR aims to make **Annoy** slightly more friendly to use inside other C++ libraries via CMake.

The first change is to add a `CMakeLists.txt` file that enables downstream projects to easily use **Annoy** with:

```
include(FetchContent)
FetchContent_Declare(
  Annoy 
  GIT_REPOSITORY https://github.com/LTLA/Annoy # replace with spotify/annoy
  GIT_TAG cmake # replace with tag name
)
FetchContent_MakeAvailable(Annoy)
target_link_libraries(mylib INTERFACE Annoy)
```

No need to vendor the files, they'll get pulled in during the build configuration of `mylib`. Usage in source files is then as simple as:

```cpp
#include "annoy/annoylib.h"

Annoy::AnnoyIndex< ... >; // etc, etc.
```

The CMake build configuration shifts the header files into an `include/annoy` subdirectory, so as to minimize conflicts with header files of the same name in other parts of the `mylib` project. Which brings me to the next point...

The second change is to add `Annoy::` namespaces to as many things as I could. Every function and class in `annoylib.h` and `kissrandom.h`. (I didn't have the courage to do so for `mman.h`; given that was mimicking a system header, I just left it alone.) The motivation is to avoid name clashes with other classes in a large project, e.g., `Base` is a particularly common name. If this PR is accepted, it would be a breaking change for anyone who is living at **Annoy**'s head. The quick-and-dirty workaround is simple; just add `using namespace Annoy;` somewhere in the source file, like I did in `annoymodule.cc`.

On a related note, the third change is to prefix as many of the newly `#define`d macros with `annoylib_` (function-like) or `ANNOYLIB_` (object-like). Again, this aims to avoid clobbering existing macros in a given compilation unit whenever somebody `#include`s the **Annoy** headers. It seems that most of these macros are only used internally, and the only observable breaking changes would be for downstream applications relying on the macro names. Not sure if that's a real problem or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,20 @@
+cmake_minimum_required(VERSION 3.14)
 project(Annoy
     DESCRIPTION "Approximate Nearest Neighbors Oh Yeah"
     VERSION 1.17.0
     LANGUAGES CXX
 )
-cmake_minimum_required(VERSION 3.14)
 
 add_library(Annoy INTERFACE)
 
-target_include_directories(Annoy INTERFACE src/)
+set(ANNOY_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include/annoy)
+file(MAKE_DIRECTORY ${ANNOY_INCLUDE_DIR})
+foreach (HEADER annoylib.h kissrandom.h mman.h)
+    file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/src/${HEADER} 
+         DESTINATION ${ANNOY_INCLUDE_DIR}/${HEADER})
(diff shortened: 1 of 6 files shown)


erikbern:
Nice – happy to merge this if this pull request is ready to be merged?

LTLA:
Yep, I'm happy with it. Note that I didn't check the Go and Lua bindings, so that might benefit from some eyes.

erikbern:
well tests are passing so let's merge :)

erikbern:
thanks!

PR #564, opened 2021-07-26
Fix namespace in go module and tests due to added Annoy namespace.

bgokden:
Compiling go module with Swig 4.0.2 creates this error:

```bash
$ swig -go -intgosize 64 -cgo -c++ src/annoygomodule.i

src/annoygomodule.h:4: Error: Nothing known about namespace 'Annoy'
```

It is fixed by explicitly defining namespace in `src/annoygomodule.i`

There was an unused parameter `info` in `test/annoy_test.go`, it is fixed to add a usage similar to other tests.

### DIFF
--- a/src/annoygomodule.i
+++ b/src/annoygomodule.i
@@ -1,5 +1,7 @@
 %module annoyindex
 
+namespace Annoy {}
+
 %{
 #include "annoygomodule.h"
 %}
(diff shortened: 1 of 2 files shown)


anukaal (approved):

erikbern:
I guess we need to port the CI code from Travis to Github Actions – a bit nervous to merge this without a green build. I'll do it this time though :)

PR #573, opened 2021-09-12
Enable direct return of <distance, index> pairs.

LTLA:
This just exposes the <distance, index> pairs for direct consumption by downstream applications. Otherwise I would have to do a round-trip through two intermediary vectors to get the vector of paired values that I wanted all along.

Impact on existing code should be minimal, the new methods are just overloads.

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -1132,12 +1132,7 @@ public:
   vector<pair<T, S> > get_nns_by_item(S item, size_t n, int search_k, bool normalize = true) const {
     vector<pair<T, S> > output;
     const Node* m = _get(item);
-    _get_all_nns(m->v, n, search_k, output);
-    if (normalize) {
-      for (size_t i = 0; i < output.size(); i++) {
-        output[i].first = D::normalized_distance(output[i].first);
-      }
-    }
+    _get_all_nns(m->v, n, search_k, output, normalize);
     return output;
   }
 
@@ -1147,12 +1142,7 @@ public:
 
   vector<pair<T, S> > get_nns_by_vector(const T* w, size_t n, int search_k, bool normalize = true) const {
     vector<pair<T, S> > output;
-    _get_all_nns(w, n, search_k, output);
-    if (normalize) {
-      for (size_t i = 0; i < output.size(); i++) {
-        output[i].first = D::normalized_distance(output[i].first);
-      }
-    }
+    _get_all_nns(w, n, search_k, output, normalize);
     return output;
   }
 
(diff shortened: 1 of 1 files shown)


erikbern:
Isn't the normalized distance just the square root of the distance or similar? Why can't you just normalize it after calling the method?

Could this be implemented in a way where you don't overload? Might be less code.

LTLA:
> Isn't the normalized distance just the square root of the distance or similar? Why can't you just normalize it after calling the method?

Depends on what the distance metric is, I suppose. Certainly we could let that responsibility fall on the user of the Annoy library, but the current `get_nns_by_item` method already takes care of normalization for us when pushing to `distances`, so it would be consistent to continue that behavior for the proposed method. You could argue that it is overkill to allow `normalize=false`; I threw that in just in case people wanted to squeeze out more efficiency when only the indices are of interest, effectively providing the same behavior as `distances = NULL` (with respect to skipping the normalization).

> Could this be implemented in a way where you don't overload? Might be less code.

What do you have in mind? I don't think the overloads really add much code, we'd have to write a new method anyway as the return values are different from the existing set of methods. The main question is whether you feel overloading is confusing - an opinion that I would not be unsympathetic to - in which case, we can always slap some other names on the proposed methods.

erikbern:
> What do you have in mind? I don't think the overloads really add much code, we'd have to write a new method anyway as the return values are different from the existing set of methods. The main question is whether you feel overloading is confusing - an opinion that I would not be unsympathetic to - in which case, we can always slap some other names on the proposed methods.

I don't fully understand this – why do the return types have to be different? The existing methods use two pointers to vectors – one for the items and one for the distances. You should be able to just add normalization to the latter without having to add a new method (just adding a bool) I think? Maybe I'm missing something

erikbern:
Oh I think I understand now. The point isn't to offer normalized vs unnormalized methods as much as returning a vector of pairs.

I'm not very convinced this is necessary. There might be a few cases where the caller wants exactly a `vector<pair<int, float> >` but that feels rare enough (and complexity of the caller building that vector simple enough) that the extra code in annoy doesn't seem worth it IMO.

erikbern:
Closing this for now, feel free to try to give a compelling case for why I'm wrong and I might open it again :)

LTLA:
That's fair enough. It's a shame I can't get the pairs directly, but the cost of allocating the two extra vectors should be minimal, so I'll just do that for now. FWIW I think my compiler does NRVO and skips the extra copy constructor when it returns the value directly in these cases; this is based on tracking the calls to the copy constructor in another project. Don't know whether that's a common optimization, though.

Issue #619, opened 2022-12-05
Dot index recall issue on last.fm dataset


pkorobov:
Hi!

I had pretty low recall for dot index on my data. Then I decided to test the [lastfm benchmark](http://ann-benchmarks.com/lastfm-64-dot_10_angular.html) manually.

Then I realized that this benchmark is actually for an angular index with manually computed 65th component -- sqrt(max_norm^2 - ||x||^2).
I've reproduced results for an angular index, but for a dot index (with the 65th being deleted) recall was again pretty low!

There's [the notebook with calculations](https://colab.research.google.com/drive/1qAspR6JPKdttqjQnAhu9_pk8fKZJBsQB?usp=sharing).

What's going on in the notebook:

1. Calculate exact dot products using brute force, then calculate top candidates
2. Construct an angular index with the additional component sqrt(max_norm^2 - ||x||^2), compute recalls for various search_k's
3. Construct a dot index without the additional component sqrt(max_norm^2 - ||x||^2), again compute recalls for various search_k's

The results are as follows (10 trees):

| search_k                  |    1000  |  10000  | 50000  |
|----------------------------|------------|-----------|-----------|
| Angular index recall (65 components) | 0.563    | 0.736    | 0.783  |
| Dot index recall (64 components)       | 0.003    | 0.034    | 0.175 |

But, as I understand, results should be at least comparable.

May there be a bug? 
If my results are correct, I would be glad to contribute with some fixes. However, by now I don't see what can be wrong in the source code of dot index.

pkorobov:
Some thoughts:

1. This is a bit strange that `two_means` doesn't use `dot_factor` at all. I don't know if it is intended to be so, but it seems to me that `dot_factor` should be involved into computations in the same way as any other component in `node->v`. At least this would be equivalent to adding the extra component and then just runnning angular.
2. The same is about the `distance` method.
3. Dot tests seem to be weak, in my opinion. I would add some accuracy tests for dot index in PR.
4. Also, please, can someone explain why there is dot_product^2 term [here](https://github.com/spotify/annoy/blob/1bdf0d5ab8c4a1f7005b7fef2c6db2d66cfb4ab6/src/annoylib.h#L584) in `margin`.
As I see it, we have a tree node with a plane defined by vector (n_1, ..., n_f, dot_factor) and a query vector q = (q_1, ..., q_f, 0), therefore margin based on them is just a dot product of first f components of them, isn't it?
However, removing the dot_factor^2 term hurts the performance.

[...]

PR #628, opened 2023-01-19
Fix incorrect dot_factor usage

pkorobov:
Annoy surprisingly didn't use `dot_factor` during `DotProduct` index construction. Here is my attempt to fix it.

I tried to reproduce the behavior of algorithm when we manually add (f+1)'th component to the data, as this is made in ann benchmarks (more information in the issue #619).

The list of fixes includes the following:
1. I made an alternative `two_means` method called `two_means_dot` which uses `dot_factor`. Maybe not the neatest solution, I'm open to improve it.
2. Fixed the formula in `margin` method of `DotProduct` class. 

`margin(a, b) = dot(a[:f], b[:f]) + a->dot_factor * b->dot_factor `, but [the code was a bit different](https://github.com/spotify/annoy/compare/master...pkorobov:fix-dot-recall?expand=1#diff-828f69d7cfead3f97fbae746c1221b43a31b62bcb960fc3f4a1782307fc5e367L584).

Note that during index construction we **need to use** `dot_factor`, therefore I added extra `margin` and `side` methods with different signatures. They take a `Node` (not an array) as the second input vector and we'll call them [while making a tree](https://github.com/spotify/annoy/compare/master...pkorobov:fix-dot-recall?expand=1#diff-828f69d7cfead3f97fbae746c1221b43a31b62bcb960fc3f4a1782307fc5e367R1376).

But during inference we don't need `dot_factor`, as the input vector must have it being equal to 0.
So during the inference the `side` / `margin` methods with old signature will be used.

3. I've changed [distance calculation](https://github.com/spotify/annoy/compare/master...pkorobov:fix-dot-recall?expand=1#diff-828f69d7cfead3f97fbae746c1221b43a31b62bcb960fc3f4a1782307fc5e367L541)

According to [the xbox paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/XboxInnerProduct.pdf), we should use Angular or L2 distance with the extra component.
I fixed this according to that.

---

After fixes runs on a lastfm dataset show the following:
| search_k                  |    1000  |  10000  | 50000  |
|----------------------------|------------|-----------|-----------|
| Angular index recall (65 components) | 0.563    | 0.736    | 0.783  |
| Dot index (master, 64 components)       | 0.003    | 0.034    | 0.175 |
| Dot index (my fix, 64 components)       | 0.514 | 0.692 | 0.787 |

So the results on this dataset finally became comparable.

Note that dot tests are passed, except the ones that check distance.
Of course, after all, I've changed it! :)
I am open to discuss how would it be better to fix them. If we really need exact dot scores, it would be better, I guess, to add some extra method to compute exactly dot scores.
But, as I think, it is better to use angular distance to form more proper splits and search for neighbors.

Also I would add an accuracy test for the lastfm dataset with only 64 components.
@erikbern may I ask you to load such a dataset with a link [like here](https://github.com/spotify/annoy/blob/1bdf0d5ab8c4a1f7005b7fef2c6db2d66cfb4ab6/test/accuracy_test.py#L31)?
Then I would shortly add one more test to `accuracy_test.py`.


pkorobov:
Removed `two_means_dot`, must look better now.

erikbern:
There's a dot product test in https://github.com/erikbern/ann-benchmarks/ that you could use if you want to.

This looks great – I'll merge this and will publish a new version to PyPI. Let me know when it's ready!

pkorobov:
Added extra accuracy tests for `lastfm-64-dot`: angular test for this dataset as is and also dot test for shortened vectors. Had to complicate code a bit to override a metric written in the dataset, but, I hope, not too much.

Also I managed to return dot products as distances and fix 4 broken tests.
I added a flag `built` to nodes. When an index is built, index item nodes will remember it.
Angular distance is not needed after the index is already constructed, so when we calculate distance between some node and an index item node we can just return dot products.

We only need order of a metric during the inference to sort NNs, and both angular distance and dot make the same order because of our index construction.

Hope, this solution looks good.

erikbern:
It doesn't change anything for the other distance metrics right? (angular etc)

pkorobov:
No, it doesn't. Particularly angular performance didn't change at all. Didn't check other metrics exact scores, but tests seem to be fine.

I only add some new methods for other metrics that don't actually do something new (they do for DotProduct). These are `update_mean` and extra `side` and `margin` methods.

pkorobov:
@erikbern I've checked the exact recalls in `accuracy_test`:

master:
fashion-mnist-784-euclidean accuracy: 99.72% (expected 90.00%)
glove-25-angular accuracy: 95.72% (expected 69.00%)
lastfm-65-angular accuracy: 66.80% (expected 60.00%)
lastfm-64-dot accuracy:  **3.69%** (expected 60.00%)
nytimes-16-angular accuracy: 98.12% (expected 80.00%)

PR:
fashion-mnist-784-euclidean accuracy: 99.72% (expected 90.00%)
glove-25-angular accuracy: 95.72% (expected 69.00%)
lastfm-65-angular accuracy: 66.80% (expected 60.00%)
lastfm-64-dot accuracy: **68.61%** (expected 60.00%)
nytimes-16-angular accuracy: 98.12% (expected 80.00%)

A little note:
lastfm-65-angular is actually the dataset from ANN benchmarks.
lastfm-64-dot is a new one for dot metric and 64-length vectors.
So naming of datasets here means exactly their contents.
I just needed to save indices with different names and these names during the test, I guess, add a bit more clarity.

Is it good to merge?

pkorobov:
@erikbern a gentle reminder that this one is waiting for review, PTAL

erikbern:
I think it would be great if @psobot could look at this! Since he's the author of the dot product code

erikbern:
Sorry for also dropping the ball on this. Let's try to get this merged. Let me see how bad the conflicts are.

erikbern:
Will merge this if tests are passing!

pkorobov:
This is great to see updates on the PR, thanks!
It seems that the [lastfm-dot-64.hdf5](http://ann-benchmarks.com/lastfm-64-dot.hdf5) dataset is not available on ann benchmarks anymore.

erikbern:
Cool let me try to fix

erikbern:
Thanks!!

Issue #633, opened 2023-02-28
Memory Leak in Annoy (get_nns_by_vector)?


siva-tonita:
The following simple Python code that loads about 100K points in 2D and performs an NN lookup in a loop (using a fixed query vector) with respect to this database seems to lead to memory leaks (memory usage keeps going up until the OS kills it).  This is a stripped down version from an actual use case (hence the artificial query point that repeats).

It will be great to have this fixed!  Thanks in advance.
--Siva

```
import random

import psutil
from annoy import AnnoyIndex

annoy_index = AnnoyIndex(f=2, metric="euclidean")

center_lat = 42.8962176
center_lng = -78.9344823

N = 100000
for idx in range(N):
    lat = center_lat + random.random() * 0.1
    lng = center_lat + random.random() * 0.1
    annoy_index.add_item(idx, [lat, lng])

print(f"Building Annoy Index", end="")
annoy_num_trees = 10
annoy_index.build(n_trees=annoy_num_trees)
print(f"Done.")

top_n = 256
avg_num_neighbors = 0.0
trials = 50000

for t in range(trials):

    if t % 1000 == 0:
        ram_bytes_used = psutil.virtual_memory()[3]
        ram_gb_used = ram_bytes_used / (2**30)
        print(f"Trial {t}. RAM Used (GB): {ram_gb_used:.4f}")

    query = [center_lat, center_lng]
    indices, distances = annoy_index.get_nns_by_vector(
        query, n=top_n, include_distances=True
    )

    del indices
    del distances
```

erikbern:
Interesting! I'm surprised this happens since none of that code has been touched for years and it feels like it should have been discovered a long time ago. Do you have any more information that's relevant, like what Python version and what OS you're using?

siva-tonita:
Yes, very surprising!
Tested on a bunch of Debian VMs with 3+ GB of RAM: Debian GNU/Linux 11 (Bullseye), Python 3.9.16
(garden variety VM from GCP).

siva-tonita:
Also:
```
$ uname -r
5.10.0-21-cloud-amd64
```
and
```
$ pip3 freeze | fgrep annoy
annoy==1.17.1
```

SteliosGian:
Looks like the issue does not exist in annoy==1.16.3

wyolland:
I can confirm the data leak on my end also
```
$ uname -r
5.4.0-146-generic
```
Same library version. Downgrading to 1.16.3 also fixed for me.

[...]

PR #639, opened 2023-04-09
Decref to object we got

erikbern:
looking at the issue in https://github.com/spotify/annoy/issues/633 and this is a wild guess

I don't fully understand the rules for reference counts in Python


erikbern:
lol i don't even know how to run the tests anymore

time to switch to pytest i guess

erikbern:
actually not sure if this is the issue – reading https://pythonextensionpatterns.readthedocs.io/en/latest/parsing_arguments.html and it seems to indicate we shouldn't have to decref here

erikbern:
rebased on latest main which runs pytest & github actions

I haven't been able to reproduce this, but assuming this doesn't segfault or similar, I think the added decrefs are correct and I'll merge it

Issue #626, opened 2023-01-15
Error when trying to install on mac 13.1, intel cpu with python 3.11.1


dk8827:
I am trying to install annoy on mac os 13.1 with Python 3.11.1
`pip3 install --user annoy`

and i get this error:

Collecting annoy
  Using cached annoy-1.17.1.tar.gz (647 kB)
  Preparing metadata (setup.py) ... done
Installing collected packages: annoy
  DEPRECATION: annoy is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
  Running setup.py install for annoy ... error
  error: subprocess-exited-with-error
  
  × Running setup.py install for annoy did not run successfully.
  │ exit code: 1
  ╰─> [21 lines of output]
      /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
        warnings.warn(
      WARNING: The wheel package is not available.
      running install
      /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
        warnings.warn(
      running build
      running build_py
      creating build
      creating build/lib.macosx-10.9-universal2-cpython-311
      creating build/lib.macosx-10.9-universal2-cpython-311/annoy
      copying annoy/__init__.py -> build/lib.macosx-10.9-universal2-cpython-311/annoy
      copying annoy/__init__.pyi -> build/lib.macosx-10.9-universal2-cpython-311/annoy
      copying annoy/py.typed -> build/lib.macosx-10.9-universal2-cpython-311/annoy
      running build_ext
      building 'annoy.annoylib' extension
      creating build/temp.macosx-10.9-universal2-cpython-311
      creating build/temp.macosx-10.9-universal2-cpython-311/src
[...]

PR #645, opened 2023-05-02
fix(#626): do NOT apply -march-native on OSX intel

doublethefish:
Simply adds an extra guard when applying `-march=native`


erikbern:
Nice thanks!

PR #649, opened 2023-05-27
Fixed compilation error with Visual Studio

jrade:
When compiling with Visual Studio one gets the error

  C:\Libraries\github\spotify\annoy\src\annoylib.h(1070,11): error C4996: 'open':
  The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _open.

The functions open(), ... are not supported by MSVS 2019. Note that they are defined by the POSIX standard, not by the C++ standard. However, MSVS 2019 provides POSIX compatible functions named _open(), ... See also: https://learn.microsoft.com/en-us/cpp/c-runtime-library/compatibility?view=msvc-170

### DIFF
--- a/src/annoylib.h
+++ b/src/annoylib.h
@@ -935,7 +935,7 @@ public:
 #ifndef _MSC_VER
     _fd = open(file, O_RDWR | O_CREAT | O_TRUNC, (int) 0600);
 #else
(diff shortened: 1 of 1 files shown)


erikbern:
Thanks!

PR #667, opened 2024-07-28
cleaned Go support and added memory leak fix

AlisaLC:
added the memory leak fix from [annoy-go](https://github.com/Rikanishu/annoy-go) implementation.
changed package name of experimental support of annoy for Go (changed from `annoyindex` to `github.com/spotify/annoy`) and reformatted tests.
it's a slightly more stable and cleaner version of the previous version.


erikbern:
I can't quite judge this but hopefully it's good :)

Issue #682, opened 2025-06-04
Getting only 1 nearest neighbor even when n is set to higher values (reopen)


Lavi101:
Issue #679 mentioned this problem. Several people are still encountering this problem, myself included. Moreover, running the provided example code of the repo reproduces only one neighbor. Changing search_k/number of trees/number of indexed points
/etc. did not help either. Does anyone know what might be the issue?

The issue is with both `get_nns_by_item` and `get_nns_by_vector` return only one neighbor (always [1]).

Running on macOS (LTS) with Python 3.13.3

Tks! :)

EvX57:
I'm experiencing the exact same issue. `get_nns_by_item()` always returns only one neighbor ([0]), regardless of the input data or hyper-parameter values.

I'm using macOS 15.5, Python 3.11.11, and annoy 1.17.3.

Any troubleshooting advice would be greatly appreciated. Thank you in advance!

caseybasichis:
Just encountered this...  

1 nearest neighbor even with the simplest tests -- this code is definitely broken.

It works fine on a slightly older machine with Python 3.11.6 instead of 3.12.9.

Is this package just abandoned?  What alternatives are you all using?

mvinyard:
@caseybasichis @EvX57 @Lavi101 thank you for reporting your experiences: this was helpful to me. After several years of using and relying on this package, I am coming to find the same experience. My best guess is an unaccounted-for dependency update. I tried in two environments using python 3.11.13 and python 3.13, to no avail. However, it does work in python 3.10.18.  am personally planning on migrating to their newer ANN library, [Voyager](https://spotify.github.io/voyager/python/).

elanou:
May be related to issue with latest clang version here: https://github.com/spotify/annoy/issues/686

Issue #679, opened 2025-04-29
Getting only 1 nearest neighbor even when n is set to higher values


sunkadshreyas:
I am referring to the below simple script. I have set n=5 and I am expecting length of results and distances to be 5. However I am getting only 1 length results and distances array. I have tried with various n and search_k values but all of them give only 1 result. The index contains n_items as expected. 
```
import numpy as np
from annoy import AnnoyIndex

# Create a simple test dataset
dim = 10
n_items = 100
vectors = np.random.random((n_items, dim)).astype('float32')

# Build an index
index = AnnoyIndex(dim, 'euclidean')
for i, v in enumerate(vectors):
    index.add_item(i, v.tolist())
index.build(10)
print(index.get_n_items())

# Test search
query = vectors[0].tolist()  
print(f"Query vector: {query}")
results, distances = index.get_nns_by_vector(query, n=5, search_k=-1, include_distances=True)
print(f"Results: {results}")
print(f"Distances: {distances}")
```

erikbern:
Try setting `search_k` to a large number!

sunkadshreyas:
> Try setting `search_k` to a large number!

Did this as well. Went upto 5000. But still getting only 1 result as part of output.

aarmey:
I can reproduce this with either `search_k=-1` or `search_k=1000`, but haven't seen this previously. Is there a dependency that might have changed?

PierreGratier:
I'm getting the same problem: starting from a clean virtual environment and  installing via `pip install annoy`, the code demo in the README

```from annoy import AnnoyIndex
import random

f = 40  # Length of item vector that will be indexed

t = AnnoyIndex(f, 'angular')
for i in range(1000):
    v = [random.gauss(0, 1) for z in range(f)]
    t.add_item(i, v)

t.build(10) # 10 trees
t.save('test.ann')



u = AnnoyIndex(f, 'angular')
u.load('test.ann') # super fast, will just mmap the file
print(u.get_nns_by_item(0, 1000)) # will find the 1000 nearest neighbors
```

Outputs `[0]` instead of the list of nns.

PR #690, opened 2026-01-07
Fix (#682 #679) macOS ARM64 crash and logic error with Numpy 2.x

BabaSanfour:
This PR resolves two critical issues observed on macOS Apple Silicon (M1/M2/M3) when building with Numpy 2.x:

1. SIGBUS (Bus Error) Crash:
**Issue:** The Node structures in `src/annoylib.h` rely on a memory layout where children and v (vector data) are contiguous. On ARM64, strict alignment requirements caused SIGBUS crashes when accessing these potentially unaligned structures.
**Fix:** Added `__attribute__((__packed__))` to the Node structs in  `Angular`, `DotProduct`, `Hamming`, and `Minkowski` definitions. This ensures the compiler generates safe code for unaligned memory access.

2. Incorrect Neighbor Count (Logic Error):

**Issue:** Queries were returning only 1 neighbor even when k=10 was requested. This was traced to the `-ffast-math `compiler flag, which enables non-IEEE 754 compliant optimizations. These optimizations likely interfered with Infinity or NaN handling in the priority queue comparisons or distance calculations.
Fix: Removed `-ffast-math` from the default compiler arguments in `setup.py`.

This solves the error in #682 #679 and would help some packages YingfanWang/PaCMAP#94

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ if platform.machine() == 'x86_64':
         extra_compile_args += ['-march=native',]
 
 if os.name != 'nt':
-    extra_compile_args += ['-O3', '-ffast-math', '-fno-associative-math']
+    extra_compile_args += ['-O3']
 
 # Add multithreaded build flag for all platforms using Python 3 and
 # for non-Windows Python 2 platforms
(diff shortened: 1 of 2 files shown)


lukemartinrulez-ship-it (approved):

BabaSanfour:
will the changes be merged?

